PR #165, opened 2024-02-20
Add package definition support

ring630:


SMoraisAnsys (dismissed):
I do have multiple questions:

1. Can we change the way things are written to use a more pythonic way of declaring attributes scopes ? Currently we leverage single underscore without any meaning behind it and I feel like we should only use it to declare an attribute as `protected`. Here I would rework the code to use `__pedb` and `__edb_object` in the `PackageDef` class. Since the `edb_object` is accessed from other modules, I would also add a property to get the object but without set it:
```python
    @property
    def edb_object(self):
        """EDB object"""
        return self.__edb_object
```

2. Since method `create` of PackageDef is only used in `add_package_def` would it make sense to extend the `__init__` of `PackageDef` to work with `edb_object` input or `name` ? If there is a case where the `PackageDef` must be updated, I would recommend to rename the method, or extract it out of the `PackageDef`, or change the way we do things as it complicates the API understanding.

3. Can we update the test to give them more sense ? Here we are testing that after setting an attribute with a value, this attribute has the value we provided. This could be enough but it would shine brighter if we tested that the attribute was updated in another way (not sure if that works though):
```python
        package = self.edbapp.definitions.add_package_def("package_1")
        assert package
        package.maximum_power = 1
        assert self.edbapp.definitions.package['package_1'].maximum_power == 1
```
Here I'm working with the object to set things and check that the edbapp is correctly updated afterward !

Apart from this remarks, great job here 👍

svandenb-dev (dismissed):
LGTM

SMoraisAnsys:
> @SMoraisAnsys _edb_object ->__edb_object is not a small refactoring. Need to do it in a separate PR

I do agree that it isn't. However, I think it can still be started here with `package_def.py` since it is a new file.
I can do it if you don't have the time to do the modifications.

SMoraisAnsys:
@ring630 Could you please take a look at my changes and proceed with the merge if everything is alright with you ?
I did add some docstring and I'm not sure if they are correct or not. Could you ensure that they are correct ?

SMoraisAnsys (approved):

PR #167, opened 2024-02-20
EMI Scanner enhancement

hui-zhou-a:


src/pyedb/misc/siw_feature_config/emc/net_tags.py
```diff
@@ -13,12 +13,14 @@ def __init__(self, element=None):
             self.isCritical = self._element.attrib["isCritical"]
             self.name = self._element.attrib["name"]
             self.type = self._element.attrib["type"] if "type" in self._element.attrib else None
+            self.Diffmatename = self._element.attrib["Diffmatename"] if "Diffmatename" in self._element.attrib else None
```
SMoraisAnsys, line 16: Is there a reason for these attributes to be using Camel case instead of Snake case ?

SMoraisAnsys (approved):

src/pyedb/misc/siw_feature_config/emc/net_tags.py
ring630: The attribute name will be put in xml file which is predefined as camel case.

PR #176, opened 2024-02-22
MAINT: Use ansys-edb-core dependency

SMoraisAnsys:
Publishing to pypi got canceled because we had a dependency to pyedb-core.
This package is now named ansys-edb-core and available on pypi !


svandenb-dev (approved):
LGTM

svandenb-dev (approved):

PR #179, opened 2024-02-23
MISC: Cleanup and updates after getting public

SMoraisAnsys:
Various changes after going public among:
- add project links in pyproject.toml
- add CI dependency in pyproject.toml and update github workflows
- remove nightly test of pyaedt using pyedb as dependency (pyedb is becoming a pyaedt dependency https://github.com/ansys/pyaedt/pull/4269)


ring630 (approved):
LGTM

PR #214, opened 2024-02-26
MAINT: Add python 3 7 compatibility

SMoraisAnsys:
This PR add python 3.7 as a valid python version for PyEDB. On top of that, the smoke tests have been added to increase the range of python version tested.

Since this python version has been added to allow users to install PyAEDT with version 3.7, a deprecated warning is triggered to notify the users of an incoming deprecation. This deprecated warning is tests.


svandenb-dev (approved):
LGTM

PR #216, opened 2024-02-27
cleanup unused methods and requirements

maxcapodi78:


pyproject.toml
```diff
@@ -112,19 +88,9 @@ full = [
...
-    "pyvista==0.38.0; python_version <= '3.7'",
-    "vtk==9.2.6",
-    "SRTM.py",
-    "utm",
-    "scikit-rf",
-    "openpyxl==3.1.2",
-]
-ci = [
-    "vtk-osmesa==9.2.20230527.dev0",
 ]
```
SMoraisAnsys, line 91: We are using that specific vtk package to avoid having issues with opengl32.dll.
I expect the documentation to fail when running some examples.

SMoraisAnsys: Nvm, everything is passing and it seems that our previous issues were related to the fact that we installed `pyvista` and/or `vtk`. Avoiding the use of `vtk-osmesa` is a good catch !

PR #232, opened 2024-03-05
conformal cutout void fix

svandenb-dev:
On specific cases when using conformal cutout could create convex cutout extent resulting in void artifact after clipping. 
Voids has been removed.


maxcapodi78 (dismissed):

src/pyedb/dotnet/edb.py
```diff
@@ -2232,7 +2232,8 @@ def _create_cutout_multithread(
             )
             if extent_type in ["Conforming", self.edb_api.geometry.extent_type.Conforming, 1] and extent_defeature > 0:
                 _poly = _poly.Defeature(extent_defeature)
-
+            if extent_type in ["Conforming", self.edb_api.geometry.extent_type.Conforming, 1]:
+                _poly = _poly.CreateFromArcs(_poly.GetArcData(), True)
```
SMoraisAnsys, line 2236: Can we leverage the previous if statement to avoid "duplicate" ? This would change things into something like:
```python
if extent_type in ["Conforming", self.edb_api.geometry.extent_type.Conforming, 1]:
    if extent_defeature > 0:
        _poly = _poly.Defeature(extent_defeature)
    _poly = _poly.CreateFromArcs(_poly.GetArcData(), True)
```

SMoraisAnsys (changes requested):

src/pyedb/dotnet/edb.py
svandenb-dev: agree

PR #252, opened 2024-03-12
convert_to_polygon() False return when not successful

gkorompi:


src/pyedb/dotnet/edb_core/edb_data/primitives_data.py
```diff
@@ -402,14 +402,17 @@ def convert_to_polygon(self):
 
         Returns
         -------
-        Converted polygon.
+        Converted polygon, when successful
+        Error message: Object type is not path, when not successful
```
SMoraisAnsys, line 406:
```suggestion
        bool, :class:`dotnet.edb_core.edb_data.primitives.EDBPrimitives`
            Polygon when successful, ``False`` when failed.
```

src/pyedb/dotnet/edb_core/edb_data/primitives_data.py
```diff
@@ -402,14 +402,17 @@ def convert_to_polygon(self):
...
 
         """
         if self.type == "Path":
             polygon_data = self.primitive_object.GetPolygonData()
             polygon = self._app.modeler.create_polygon(polygon_data, self.layer_name, [], self.net_name)
             self.primitive_object.Delete()
             return polygon
+        else:
+            raise AttributeError("Object type is not path")
```
SMoraisAnsys, line 415:
```suggestion
            return False
```

SMoraisAnsys (changes requested):
I like the idea of raising an exception but pyedb is mainly returning `False` when something is going wrong. I propose to be consistent for the moment and change everything at once later on.

@svandenb-dev Seems like more and more tests do not leverage the test instance `self.edbapp` and tend to create new `Edb` instances on the fly. Should be try to move to that approach for every tests ?

SMoraisAnsys (commented):
LGTM, @svandenb-dev can you approve ?

svandenb-dev:
> I like the idea of raising an exception but pyedb is mainly returning `False` when something is going wrong. I propose to be consistent for the moment and change everything at once later on.
> 
> @svandenb-dev Seems like more and more tests do not leverage the test instance `self.edbapp` and tend to create new `Edb` instances on the fly. Should be try to move to that approach for every tests ?

@SMoraisAnsys Yes indeed. The problem is that some tests are destroying the layout integrity like the cutout. I agree to move to that direction then, with making sure we got a clean database. Experience has shown in the past that reusing the same EDB was providing conflicts due to some previous tests.

SMoraisAnsys (approved):
LGTM

PR #253, opened 2024-03-13
changing from internal EDB layer to pyedb one from primitive object

svandenb-dev:
This is a test PR as layer property was returning internal EDB Layer object. Change to pyedb layer type. Small change with potential  large impact. @maxcapodi78 please review the code when you can. For some reasons the local test was failing on one multithreaded command during cutout I had to remove the thread. Let see how it goes on the github build machine.

### DIFF
--- a/tests/legacy/system/test_edb_modeler.py
+++ b/tests/legacy/system/test_edb_modeler.py
@@ -86,8 +86,8 @@ class TestClass:
 
     def test_modeler_primitives_by_layer(self):
         """Evaluate modeler primitives by layer"""
-        assert self.edbapp.modeler.primitives_by_layer["1_Top"][0].layer_name == "1_Top"
-        assert self.edbapp.modeler.primitives_by_layer["1_Top"][0].layer.GetName() == "1_Top"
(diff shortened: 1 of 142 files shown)


tests/legacy/system/test_edb_modeler.py
```diff
@@ -86,8 +86,8 @@ def test_modeler_paths(self):
 
     def test_modeler_primitives_by_layer(self):
         """Evaluate modeler primitives by layer"""
-        assert self.edbapp.modeler.primitives_by_layer["1_Top"][0].layer_name == "1_Top"
-        assert self.edbapp.modeler.primitives_by_layer["1_Top"][0].layer.GetName() == "1_Top"
+        assert self.edbapp.modeler.primitives_by_layer["1_Top"][0].layer.name == "1_Top"
+        assert self.edbapp.modeler.primitives_by_layer["1_Top"][0].layer.name == "1_Top"
```
SMoraisAnsys, line 90: Seems like this lines are duplicates

SMoraisAnsys (approved):

PR #254, opened 2024-03-13
enabling locked edb loading

svandenb-dev:
When an aedt file or lock file is detected, trying to delete those file to preventing loading EDB.


src/pyedb/dotnet/edb.py
```diff
@@ -213,6 +213,19 @@ def __init__(
...
+                    os.remove(aedt_file)
+                    self.logger.info(f"aedt file {aedt_file}, removing this file to allow loading EDB")
+                except:
+                    self.logger.error("aedt file is located at same location as EDB but failed to delete.")
+                if os.path.isfile(aedt_file + ".lock"):
+                    try:
+                        os.remove(aedt_file + ".lock")
+                        self.logger.info(f"file {aedt_file} is locked, deleting lock file before loading EDB")
+                    except:
+                        self.logger.error(f"aedt file {aedt_file} seems to be locked but failed to remove lock file")
```
SMoraisAnsys, line 228:
```suggestion
            files = [aedt_file, aedt_file + ".lock"]
            for file in files:
                if os.path.isfile(file):
                    try:
                        shutil.rmtree(self.filename)
                        self.logger.info(f"Removing {file} to allow loading EDB file.")
                    except:
                        self.logger.info(f"Failed to delete {file} which is located at the same location as the EDB file.")
```

SMoraisAnsys (changes requested):
@svandenb-dev Instead of adding the files test.aedt and test.aedt.lock, could we do some unit test where the existence of a file is patched ? I could help you in doing that if you want

SMoraisAnsys (approved):
LGTM

SMoraisAnsys:
@svandenb-dev I added a unit tests instead of the system one and cleaned up afterward. Btw there was a bug with the removed file (`self.filename` instead of `file`)

PR #259, opened 2024-03-13
unite_polygons_on_layer bug fix

gkorompi:


svandenb-dev (approved):
LGTM

gkorompi:
changed input net_list: list of nets, to net_names_list: str inputs of net names. The net_list was not working properly in the first place, so i do not think this is a problem.

SMoraisAnsys:
@gkorompi Sorry but I interrupted the CICD workflow to let the release pass first. I'll run it back later !

PR #281, opened 2024-03-23
Bug Fix layer_name

gkorompi:


src/pyedb/dotnet/edb_core/edb_data/primitives_data.py
```diff
@@ -160,13 +160,14 @@ def layer_name(self):
...
-            lay = self._core_stackup.layers["TOP"]._edb_layer
-            if lay:
-                self.primitive_object.SetLayer(lay)
+        layer_list = list(self._core_stackup.layers.keys())
+        if isinstance(val, str) and val in layer_list:
+            layer = self._core_stackup.layers[val]._edb_layer
+            if layer:
+                self.primitive_object.SetLayer(layer)
             else:
                 raise AttributeError("Layer {} not found in layer".format(val))
```
SMoraisAnsys, line 169:
```suggestion
                raise AttributeError("Layer {} not found.".format(val))
```

SMoraisAnsys (commented):
Any reason to remove `test_modeler_primitives_by_layer` ?

gkorompi:
> Any reason to remove `test_modeler_primitives_by_layer` ?

In this function on the second line: assert self.edbapp.modeler.primitives_by_layer["1_Top"][0].layer.GetName() == "1_Top"

there is an error: AttributeError: 'StackupLayerEdbClass' object has no attribute 'GetName'

I do not think this is related with the changes I made in layer_name function. If i triggered this mistake, please let me know.

src/pyedb/dotnet/edb_core/edb_data/primitives_data.py
```diff
@@ -160,13 +160,14 @@ def layer_name(self):
...
+        layer_list = list(self._core_stackup.layers.keys())
+        if isinstance(val, str) and val in layer_list:
+            layer = self._core_stackup.layers[val]._edb_layer
+            if layer:
+                self.primitive_object.SetLayer(layer)
             else:
-                raise AttributeError("Layer {} not found in layer".format(val))
-        elif isinstance(val, type(self._core_stackup.layers["TOP"])):
+                raise AttributeError("Layer {} not found.".format(val))
+        elif isinstance(val, type(self._core_stackup.layers[layer_list[0]])):
```
SMoraisAnsys, line 170: Does it make sense to change the primitive object layer when providing something else than a string ?

gkorompi: layer_list[0] is also a string. Problem is if in the stackup there is not a layer named "TOP", the function returns an error. So layer_list[0] calls the str name of the first layer in the stackup.

SMoraisAnsys: Then i would advise to refactor the code to avoid calling `isinstance(val, str)` twice (once directly, the second time through `type(self._core_stackup.layers[layer_list[0]])`).

SMoraisAnsys (changes requested):

tests/legacy/system/test_edb_modeler.py
```diff
@@ -390,3 +390,24 @@ def test_duplicate(self):
         assert edbapp.modeler.primitives_by_layer["mid_gnd"]
         assert edbapp.modeler.primitives_by_layer["bot_gnd"]
         edbapp.close()
+
+    def test_layer_name(self):
+        edbapp = Edb()
```
ring630, line 395: Please use below edb for testing. Avoid create a layout for every test.
edbapp = Edb(target_path, desktop_version)

src/pyedb/dotnet/edb_core/edb_data/primitives_data.py
gkorompi: isinstance(val, str) refers to input str name of layer: i.e. "TOP"
isinstance(val, type(self._core_stackup.layers[layer_list[0]])) refers to the actual object of layer: StackupLayerEdbClass.
So it works for both types.

tests/legacy/system/test_edb_modeler.py
gkorompi: Okay i corrected it please check

SMoraisAnsys (approved):
LGTM

PR #309, opened 2024-04-02
FIX: 289 xml for different layer names

gkorompi:


gkorompi:
the part that i deleted basically stops layers with different names than the existing stackup. Hence if the xml file has as a first layer "signal" and the edb file stackup has "top" then it returns False.

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -2107,11 +2107,6 @@ def _import_xml(self, file_path):
         for i in list(lc_import.Layers(self._pedb.edb_api.cell.layer_type_set.AllLayerSet)):
             dumy_layers[i.GetName()] = i.Clone()
 
-        for name in self.layers.keys():
-            if not name in dumy_layers:
-                logger.error("{} doesn't exist in xml".format(name))
-                return False
-
```
SMoraisAnsys, line 2114: Would it make sense to log a warning instead of removing the log of an error and return value ?

gkorompi: I think that it should check the elevation of the layers, aka the thickness, and the number of them. I changed the code in order to check based on this. @svandenb-dev please review.

svandenb-dev (changes requested):
This method has been changed in another PR and I don't this the code is valid anymore. Please wait https://github.com/ansys/pyedb/pull/321 is merged and check if this code is still applicable

PR #329, opened 2024-04-08
get primitives by layer and points

svandenb-dev:
adding new feature #

### DIFF
--- a/src/pyedb/dotnet/edb_core/layout.py
+++ b/src/pyedb/dotnet/edb_core/layout.py
@@ -241,6 +241,65 @@ class EdbLayout(object):
         return objinst
 
     @pyedb_function_handler()
+    def get_primitive_by_layer_and_point(self, point=None, layer=None, nets=None):
+        """Return primitive given coordinate point [x, y], layer name and nets.
+
+        Parameters
+        ----------
+        point : list
+            Coordinate [x, y]
+
+        layer : list or str, optional
+            list of layer name or layer name applied on filter.
+
+        nets : list or str, optional
+            list of net name or single net name applied on filter
+
+        Returns
+        -------
+        list of :class:`pyedb.dotnet.edb_core.edb_data.primitives_data.EDBPrimitives`
+            List of primitives, polygons, paths and rectangles.
+        """
+        if isinstance(layer, str):
+            if layer not in list(self._pedb.stackup.signal_layers.keys()):
+                layer = None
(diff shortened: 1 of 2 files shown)


src/pyedb/dotnet/edb_core/layout.py
```diff
@@ -240,6 +240,65 @@ def get_polygons_by_layer(self, layer_name, net_list=None):
...
+            list of net name or single net name applied on filter
+
+        Returns
+        -------
+        list of :class:`pyedb.dotnet.edb_core.edb_data.primitives_data.EDBPrimitives`
+            List of primitives, polygons, paths and rectangles.
+        """
+        if isinstance(layer, str):
+            if layer not in list(self._pedb.stackup.signal_layers.keys()):
+                layer = None
```
SMoraisAnsys, line 265:
```suggestion
        if isinstance(layer, str) and layer not in list(self._pedb.stackup.signal_layers.keys()):
            layer = None
```

src/pyedb/dotnet/edb_core/layout.py
```diff
@@ -240,6 +240,65 @@ def get_polygons_by_layer(self, layer_name, net_list=None):
...
+        if not isinstance(point, list) and len(point) == 2:
+            return False
```
SMoraisAnsys, line 267:
```suggestion
        if not isinstance(point, list) and len(point) == 2:
            self._logger.error("Provided point must be a list of two values")
            return False
```

src/pyedb/dotnet/edb_core/layout.py
```diff
@@ -240,6 +240,65 @@ def get_polygons_by_layer(self, layer_name, net_list=None):
...
+        pt = self._edb.geometry.point_data(point[0], point[1])
+        if isinstance(nets, str):
+            nets = [nets]
+        elif nets and not isinstance(nets, list) and len(nets) == len([net for net in nets if isinstance(net, str)]):
+            _nets = []
+            for net in nets:
+                if net not in self._pedb.nets:
+                    self._logger.error(
+                        f"Net {net} used to find primitive from layer point and net not found, skipping " f"net"
```
SMoraisAnsys, line 276:
```suggestion
                        f"Net {net} used to find primitive from layer point and net not found, skipping it."
```

SMoraisAnsys (changes requested):
Great job, just a couple of changes that could help with debugging.

ring630 (approved):
LGTM

PR #339, opened 2024-04-09
FEATURE: package definition support in edb configuration

ring630:


src/pyedb/dotnet/edb_core/configuration.py
```diff
@@ -678,3 +682,33 @@ def _load_padstacks(self):
...
+
+    @pyedb_function_handler
+    def _load_package_def(self):
+        """Imports package definition information from JSON."""
+        comps = self._pedb.components.components
+        for pkgd in self.data["package_definitions"]:
+            name = pkgd["name"]
+            if name in self._pedb.definitions.package:
+                self._pedb.definitions.package[name].delete()
+            from pyedb.dotnet.edb_core.definition.package_def import PackageDef
```
SMoraisAnsys, line 694: Is it possible to move that call at the top of the module or does it creates circular import ?

SMoraisAnsys (approved):
Just a minor comment, otherwise LGTM

SMoraisAnsys:
@ring630 Why did you added dotnet label ?

svandenb-dev (dismissed):
LGTM

src/pyedb/dotnet/edb_core/configuration.py
ring630: No. I will correct this.

svandenb-dev (approved):
LGTM

PR #350, opened 2024-04-10
CI: Check title with ansys action

SMoraisAnsys:
As title says.

~> [!WARNING]~
~> This branch should not be merged until a new ansys/action release is performed.~

~Note: Leverage https://github.com/ansys/actions/issues/446 (which is not released yet but part of main version)~

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check commit name
-        uses: ansys/actions/commit-style@v5
+        uses: ansys/actions/commit-style@v6
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -27,33 +27,15 @@ concurrency:
...
     runs-on: ubuntu-latest
     steps:
-      - name: Get PR title
-        env:
-          REPOSITORY_OWNER: ${{ github.repository_owner }}
-          REPOSITORY_NAME: ${{ github.event.repository.name }}
-          PULL_REQUEST_NUMBER: ${{ github.event.number }}
-        uses: actions/github-script@v7
+      - name: Check commit name
+        uses: ansys/actions/commit-style@v5
```
SMoraisAnsys, line 35:
```suggestion
        uses: ansys/actions/commit-style@v6
```

PR #356, opened 2024-04-11
FIX: via_plating_thickness

ring630:


SMoraisAnsys (commented):
LGTM, if it is related to an issue, could you reference it ?

PR #358, opened 2024-04-11
FEATURE: rlc_component without name

gkorompi:


tests/legacy/system/test_edb_modeler.py
```diff
@@ -466,7 +466,7 @@ def rlc_component_302(self):
         self.local_scratch.copyfolder(source_path_edb, target_path_edb)
         edbapp = Edb(target_path_edb, desktop_version)
         pins = edbapp.components.get_pin_from_component("C31")
-        assert edbapp.components.create_rlc_component([pins[0], pins[1]], r_value=0, component_name="TEST")
+        assert edbapp.siwave.create_rlc_component([pins[0], pins[1]])
```
SMoraisAnsys, line 469: Any reason for such change ?

gkorompi: instead of writing a new test i just added this line, which tests the changes.

SMoraisAnsys (dismissed):
LGTM

tests/legacy/system/test_edb_modeler.py
SMoraisAnsys: I see that now the two lines are kept so my previous comment does not apply any more.

SMoraisAnsys:
@gkorompi Please avoid using auto merge because I can't accept a review otherwise (my approval would make the PR getting merge). I prefer that either @ring630 or @svandenb-dev also give their opinion before merging something as they are the core maintainers.

ring630 (approved):
LGTM

PR #362, opened 2024-04-12
FIX: package def

ring630:


ring630:
@SMoraisAnsys pycharm doesn't allow ":" in PR name. Is it possible to change the naming rule from "FIX:" to "FIX-"?

SMoraisAnsys:
> @SMoraisAnsys pycharm doesn't allow ":" in PR name. Is it possible to change the naming rule from "FIX:" to "FIX-"?

Do you mean branch name ? It shouldn't be a problem to use ":" in a PR title. If that's the case, i assume there is a way to make it work. Using "-" instead is not standard commit style and is quite awkward IMO

src/pyedb/dotnet/edb_core/geometry/point_data.py
```diff
@@ -0,0 +1,12 @@
+class PointData:
+    """Point Data."""
+
+    def __init__(self, pedb, edb_object=None, x=None, y=None):
+        self._pedb = pedb
+        if edb_object:
+            self._edb_object = edb_object
+        else:
+            self._edb_object = self._pedb.edb_api.geometry.point_data(
```
SMoraisAnsys, line 9: Won't this fail if `x` or `y` is `None`?

src/pyedb/dotnet/edb_core/geometry/polygon_data.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(
+        self,
+        pedb,
+        edb_object=None,
+        create_from_points=None,
+        create_from_circle=None,
+        create_from_rectangle=None,
+        create_from_bounding_box=None,
+        **args,
```
SMoraisAnsys, line 17: Could you use `kwargs` instead here ? `args` is more commonly used like `*args`

SMoraisAnsys (changes requested):
On top of the minor comment I left, could you add the copyright on each new files ? Some do not have it.

SMoraisAnsys (approved):
LGTM, thanks for taking into account previous review comments !

PR #367, opened 2024-04-15
FEATURE: get_plot_data()

gkorompi:


src/pyedb/dotnet/edb_core/nets.py
```diff
@@ -472,7 +472,7 @@ def _get_points_for_plot(self, my_net_points):
     @pyedb_function_handler()
     def get_plot_data(
         self,
-        nets,
+        nets=None,
```
SMoraisAnsys, line 475: Could you update the description to:
- add `optional` to `nets` types (this should already have been performed)
- specify that, by default, all nets will be considered

SMoraisAnsys (changes requested):
@svandenb-dev @ring630 Can one of you confirm the changes as it will modify the default behavior of the method.

src/pyedb/dotnet/edb_core/nets.py
```diff
@@ -483,8 +483,8 @@ def get_plot_data(
 
         Parameters
         ----------
-        nets : str, list
-            Name of the net or list of nets to plot. If `None` all nets will be plotted.
+        nets : str, list, optional
+            Name of the net or list of nets to plot. If `None` all nets will be plotted. (default)
```
SMoraisAnsys, line 487:
```suggestion
            Name of the net or list of nets to plot. If `None` (default value) all nets will be plotted.
```

ring630:
> @svandenb-dev @ring630 Can one of you confirm the changes as it will modify the default behavior of the method.

@SMoraisAnsys it is a bug. When nets==None, all nets are plotted. However, nets is a mandatory input, which makes no sense.

ring630 (approved):
LGTM

SMoraisAnsys (approved):

PR #374, opened 2024-04-16
FEATURE: xml_rename fix

gkorompi:


src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -1698,8 +1698,12 @@ def _import_dict(self, json_dict, rename=False):
...
         layout_layers = list(self.stackup_layers.keys())
         renamed_layers = {}
-        if rename and len(config_file_layers) == len(layout_layers):
-            for lay_ind in range(len(list(temp.keys()))):
+        if rename and len(config_file_layers) >= len(layout_layers):
+            for lay_ind in range(len(layout_layers)):
+                if not config_file_layers[lay_ind] == layout_layers[lay_ind]:
+                    renamed_layers[layout_layers[lay_ind]] = config_file_layers[lay_ind]
+        elif rename and len(config_file_layers) < len(layout_layers):
+            for lay_ind in range(len(config_file_layers)):
```
ring630, line 1706: if rename:
     # When rename is True, Layer count must be the same.
else:
     # When rename is False. Layer count from json must not be higher than the one from edb. The names of each must be the same.
@svandenb-dev

gkorompi: i made it so that if rename: True, it will rename all layers starting from top, and if the imported layer has more layers than the existing ones, then the remaining ones are not taken into consideration when renaming. 
Is this not ok?

ring630: It is not possible the change the layer count in the mean time layer names are different.

PR #376, opened 2024-04-16
FIX: protect import from matplotlib

### LINKED ISSUE #351: Missing dependency declaration or ImportError check for matplotlib

### Description of the bug

https://github.com/ansys/pyedb/blob/c5a03f383e7f2b2bd8fc809012746fbe2cbe6d6c/src/pyedb/dotnet/edb_core/stackup.py#L36 imports from `matplotlib` outside a `try-except ImportError` block, but `matplotlib` is not in the primary dependencies list in `pyproject.toml`.

### Steps To Reproduce

1. Create a fresh virtual environment and activate it
2. `pip install pyedb` to install PyEDB with only the base dependency set
3. Run the following interaction in a Python interactive shell

```
(venv) C:\scratch>python
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyedb import Edb
>>> db = Edb(edbversion="2024.1")
C:\scratch\venv\lib\site-packages\pyedb\dotnet\edb_core\edb_data\components_data.py:36: UserWarning: The NumPy module is required to run some functionalities of EDB.
Install with

pip install numpy

Requires CPython.
  warnings.warn(
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\scratch\venv\lib\site-packages\pyedb\generic\design_types.py", line 108, in Edb
    from pyedb.dotnet.edb import Edb as app
  File "C:\scratch\venv\lib\site-packages\pyedb\dotnet\edb.py", line 87, in <module>
    from pyedb.dotnet.edb_core.stackup import Stackup
  File "C:\scratch\venv\lib\site-packages\pyedb\dotnet\edb_core\stackup.py", line 36, in <module>
    import matplotlib.colors
ModuleNotFoundError: No module named 'matplotlib'
>>>
```

Windows

3.10

### Installed packages

```
annotated-types==0.6.0
ansys-pythonnet==3.1.0rc3
cffi==1.16.0
clr-loader==0.2.6
pycparser==2.22
pydantic==2.6.4
pydantic_core==2.16.3
pyedb==0.7.1
pywin32==306
toml==0.10.2
typing_extensions==4.11.0
```

isaacansys:
Protect the import of matplotlib.colors in pyedb.dotnet.edb_core.stackup to avoid import errors when users install only the base dependency set, which does not include matplotlib.

Fixes #351

### DIFF
--- a/src/pyedb/dotnet/edb_core/stackup.py
+++ b/src/pyedb/dotnet/edb_core/stackup.py
@@ -33,8 +33,6 @@ import logging
 import math
 import warnings
 
-import matplotlib.colors
-
 from pyedb.dotnet.edb_core.edb_data.layer_data import (
     LayerEdbClass,
     StackupLayerEdbClass,
@@ -48,9 +46,15 @@ from pyedb.generic.general_methods import (
 )
 from pyedb.misc.aedtlib_personalib_install import write_pretty_xml
 
+colors = None
 pd = None
 np = None
 if not is_ironpython:
+    try:
+        import matplotlib.colors as colors
+    except ImportError:
+        colors = None
+
     try:
         import numpy as np
     except ImportError:
@@ -2130,7 +2134,7 @@ class Stackup(object):
                 layer = {"name": l.attrib["Name"]}
                 for k, v in l.attrib.items():
                     if k == "Color":
-                        layer[k.lower()] = [int(x * 255) for x in list(matplotlib.colors.to_rgb(v))]
+                        layer[k.lower()] = [int(x * 255) for x in list(colors.to_rgb(v))]
                     elif k == "Thickness":
                         layer[k.lower()] = v + length_unit
                     elif v == "conductor":
--- a/src/pyedb/dotnet/edb_core/stackup.py
+++ b/src/pyedb/dotnet/edb_core/stackup.py
@@ -2114,6 +2114,9 @@ class Stackup(object):
         bool
             ``True`` when successful, ``False`` when failed.
         """
+        if not colors:
+            self._pedb.logger.error("Matplotlib is needed. Please, install it first.")
+            return False
         tree = ET.parse(file_path)
         root = tree.getroot()
         stackup = root.find("Stackup")


src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -2130,7 +2134,7 @@ def _import_xml(self, file_path, rename=False):
                 layer = {"name": l.attrib["Name"]}
                 for k, v in l.attrib.items():
                     if k == "Color":
-                        layer[k.lower()] = [int(x * 255) for x in list(matplotlib.colors.to_rgb(v))]
+                        layer[k.lower()] = [int(x * 255) for x in list(colors.to_rgb(v))]
```
SMoraisAnsys, line 2137: If `colors` is equal to `None`, wouldn't that LOC raise an exception ?

isaacansys: Yes, but here I am following what was done in this module for numpy and pandas as well.  These cannot be imported in IronPython and are not in the base dependency set so using functions that depend on them will fail unless you install pyedb[full].

Open to suggestions on how to better handle this, but this is how it is being done for the other two moduels currently.

isaacansys: Ah, I see the check for np is None before it is used, I'll add this.

isaacansys: @SMoraisAnsys I've added a check for colors being None to gracefully fail in this case, in the same way as is done if np is None elsewhere.

SMoraisAnsys: Yeah, that will avoid the problem mentioned. Thanks for handling it :)

SMoraisAnsys (approved):
LGTM, thanks for taking into account previous review !

svandenb-dev (approved):
LGTM

PR #391, opened 2024-04-19
FEATURE: add types create_pingroup_from_pins

gkorompi:


src/pyedb/dotnet/edb_core/components.py
```diff
@@ -858,7 +858,7 @@ def create_port_on_component(
         self,
         component,
         net_list,
-        port_type=SourceType.CoaxPort,
+        port_type=SourceType.CircPort,
```
ring630, line 861: The default value should be not changed.

src/pyedb/dotnet/edb_core/components.py
```diff
@@ -1335,18 +1335,21 @@ def add_rlc_boundary(self, component=None, circuit_type=True):
             return True
 
     @pyedb_function_handler()
-    def _create_pin_group_terminal(self, pingroup, isref=False, term_name=None):
+    def _create_pin_group_terminal(self, pingroup, isref=False, term_name=None, term_type="gap"):
```
ring630, line 1338: Is gap port supported on a pin grroup?

ring630 (changes requested):

src/pyedb/dotnet/edb_core/components.py
gkorompi: Yes. The port created when running this function PinGroupTerminal.Create(...) is a gap port. I added the other options of changing it to circuit port and auto.

svandenb-dev: I wouldn't set "gap" as default. Most of users are dealing with circuit ports

src/pyedb/dotnet/edb_core/components.py
```diff
@@ -1360,6 +1363,10 @@ def _create_pin_group_terminal(self, pingroup, isref=False, term_name=None):
         pingroup_term = self._edb.cell.terminal.PinGroupTerminal.Create(
             self._active_layout, pingroup.GetNet(), term_name, pingroup, isref
         )
+        if term_type == "circuit":
+            pingroup_term.SetIsCircuitPort(True)
+        elif term_type == "auto":
```
ring630, line 1368: There is no need to have term_type argument. Just hardcode it to circuit port.

ring630 (approved):

PR #407, opened 2024-04-24
FIX: return error message when trying to create_ports_on_pins of RLC comp

gkorompi:


src/pyedb/dotnet/edb_core/components.py
```diff
@@ -785,6 +785,9 @@ def create_port_on_pins(self, refdes, pins, reference_pins, impedance=50.0, port
             reference_pins = [reference_pins]
         if isinstance(refdes, str) or isinstance(refdes, EDBComponent):
             refdes = self.instances[refdes]
+        if refdes.rlc_values:
```
ring630, line 788: It is better to check type rathen than rlc value. Why is RLC not supported?

gkorompi: Yes but type is EDBComponent for all inputs (even inductors, caps etc). Ports are not successfully created on RLC pins. Should this be added for 3DL?

svandenb-dev: @gkorompi there is a command to replace RLC by ports which deactivate component and create a circuit port. You might want to call this one when you hit such component ? 
Also also rather component type instead of rlc_value.

src/pyedb/dotnet/edb_core/components.py
```diff
@@ -785,6 +785,10 @@ def create_port_on_pins(self, refdes, pins, reference_pins, impedance=50.0, port
             reference_pins = [reference_pins]
         if isinstance(refdes, str) or isinstance(refdes, EDBComponent):
             refdes = self.instances[refdes]
+        for i in refdes.rlc_values:
+            if i:
+                self._logger.error("Ports cannot be assigned on RLC components")
+                return False
```
SMoraisAnsys, line 791:
```suggestion
        if any(refdes.rlc_values):
            self._logger.error("Ports cannot be assigned on RLC components")
            return False
```

SMoraisAnsys (commented):
@svandenb-dev @ring630 Any feedback on this PR state ?

ring630:
> @svandenb-dev @ring630 Any new feedback on this PR state ?

I am ok with pr, even though this method is wired to me.

PR #413, opened 2024-04-24
FIX: Don't always remove AEDT project files

### LINKED ISSUE #412: Opening an EDB should not (always?) delete corresponding aedt project file

### Description of the bug

#254 introduces an issue where opening an EDB with a corresponding aedt project file will always attempt to delete the project file; this is not always desired e.g. when opening the EDB during an automated test to confirm that earlier actions produced the correct result.  At a minimum when an EDB is opened read-only the aedt (and aedt.lock) file(s) should be left in place if they exist.

There is also a bug with the implementation of these deletions: shutil.rmtree cannot be used to delete a file (at least in Python 3.10) as it throws a NonDirectoryException in this case.  So currently this is a no-op, but before this is fixed I'd like to open a discussion as above on whether the aedt file should always be deleted or not.

### Steps To Reproduce

Run the following script in a virtual environment with both pyaedt and pyedb installed:

```
from pathlib import Path
import shutil
import tempfile

from pyaedt import Hfss3dLayout
from pyedb import Edb


version = "2024.1"
folder = Path(tempfile.mkdtemp())
aedt_path = folder / "project.aedt"

with Hfss3dLayout(close_on_exit=True, non_graphical=True, specified_version=version) as hfss:
    hfss.save_project(aedt_path.as_posix())

assert aedt_path.is_file()

edb_path = aedt_path.with_suffix(".aedb")

assert edb_path.is_dir()

db = Edb(edb_path.as_posix(), edbversion=version, isreadonly=True)
try:
    pass
finally:
    db.close_edb()

assert aedt_path.is_file()
```

Initially (with current main branch) there will be no assertion error because https://github.com/ansys/pyedb/blame/d135483cd94cb87d8c27c2ae67de302718430f4b/src/pyedb/dotnet/edb.py#L221 will throw a NonDirectoryException.  Changing this line to os.unlink(file) will remove the file but then result in an AssertionError on line 28 of the script.

Windows

3.10

### Installed packages

```
annotated-types==0.6.0
ansys-pythonnet==3.1.0rc3
attrs==23.2.0
cffi==1.16.0
cfgv==3.4.0
clr-loader==0.2.6
colorama==0.4.6
contourpy==1.2.1
coverage==7.5.0
cycler==0.12.1
defusedxml==0.7.1
distlib==0.3.8
exceptiongroup==1.2.0
execnet==2.1.1
filelock==3.13.4
fonttools==4.51.0
fpdf2==2.7.8
identify==2.5.35
iniconfig==2.0.0
jsonschema==4.21.1
jsonschema-specifications==2023.12.1
kiwisolver==1.4.5
matplotlib==3.8.4
mock==5.1.0
nodeenv==1.8.0
numpy==1.26.4
packaging==24.0
pandas==2.2.2
pillow==10.3.0
platformdirs==4.2.0
pluggy==1.4.0
plumbum==1.8.2
pre-commit==3.7.0
psutil==5.9.8
pyaedt==0.8.9
pycparser==2.22
pydantic==2.7.0
pydantic_core==2.18.1
-e git+https://github.com/ansys/pyedb@0f0518d753f7bfb828a9f3cee9d59565ac963f21#egg=pyedb
pyparsing==3.1.2
pytest==8.1.1
pytest-cov==5.0.0
pytest-xdist==3.5.0
python-dateutil==2.9.0.post0
pytomlpp==1.0.13
pytz==2024.1
pywin32==306
PyYAML==6.0.1
referencing==0.35.0
rpds-py==0.18.0
rpyc==6.0.0
six==1.16.0
toml==0.10.2
tomli==2.0.1
typing_extensions==4.11.0
tzdata==2024.1
virtualenv==20.25.1
```

isaacansys:
Remove AEDT project-related files on EDB open (e.g. file.aedt and/or file.aedt.lock for file.aedb) only if Edb constructor is run with isreadonly False and remove_existing_aedt True.

Fixes #412

### DIFF
--- a/src/pyedb/dotnet/edb.py
+++ b/src/pyedb/dotnet/edb.py
@@ -174,15 +174,16 @@ class Edb(Database):
 
     def __init__(
         self,
-        edbpath=None,
-        cellname=None,
-        isreadonly=False,
-        edbversion=None,
-        isaedtowned=False,
+        edbpath: str = None,
+        cellname: str = None,
+        isreadonly: bool = False,
+        edbversion: str = None,
+        isaedtowned: bool = False,
         oproject=None,
-        student_version=False,
-        use_ppe=False,
-        technology_file=None,
+        student_version: bool = False,
+        use_ppe: bool = False,
+        technology_file: str = None,
+        remove_existing_aedt: bool = False,
     ):
         edbversion = get_string_version(edbversion)
         self._clean_variables()
@@ -213,17 +214,8 @@ class Edb(Database):
                 os.path.dirname(edbpath),
                 "pyedb_" + os.path.splitext(os.path.split(edbpath)[-1])[0] + ".log",
             )
-            aedt_file = os.path.splitext(edbpath)[0] + ".aedt"
-            files = [aedt_file, aedt_file + ".lock"]
-            for file in files:
-                if os.path.isfile(file):
-                    try:
-                        shutil.rmtree(file)
-                        self.logger.info(f"Removing {file} to allow loading EDB file.")
-                    except:
-                        self.logger.info(
-                            f"Failed to delete {file} which is located at the same location as the EDB file."
-                        )
+            if not isreadonly:
+                self._check_remove_project_files(edbpath, remove_existing_aedt)
 
         if isaedtowned and (inside_desktop or settings.remote_rpc_session):
             self.open_edb_inside_aedt()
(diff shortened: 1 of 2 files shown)


isaacansys:
@svandenb-dev BTW it looks like PyAEDT already has some logic to handle an existing .aedt file and open it if you try to open an edb.def or .aedb folder: https://github.com/ansys/pyaedt/blob/87739a9ea3223ee8e57d3d7c6ca7569a4f18eefe/pyaedt/application/Design.py#L1122

I guess we should leave it to users to understand when they have to remove the AEDT file before opening a modified EDB (with the warning I'm introducing here), what do you think?

svandenb-dev (dismissed):
LGTM

SMoraisAnsys (approved):
LGTM, thanks for doing the requested changes !

PR #421, opened 2024-04-26
FEAT: Broadband simulation setup support with SimulationConfiguration class

svandenb-dev:
Adding support of adaptive broadband with SimulationConfiguration and Json file.


ring630 (approved):

src/pyedb/dotnet/edb_core/edb_data/simulation_configuration.py
```diff
@@ -1902,6 +1906,56 @@ def mesh_sizefactor(self, value):
...
+
+    @adaptive_type.setter
+    def adaptive_type(self, value):
+        if isinstance(value, int):
+            if value == 0:
+                self._adaptive_type = AdaptiveType.SingleFrequency
+            elif value == 1:
+                self._adaptive_type = AdaptiveType.MultiFrequency
+            elif value == 2:
+                self._adaptive_type = AdaptiveType.BroadBand
```
SMoraisAnsys, line 1927:
```suggestion
        if isinstance(value, int) and value in range(3):
            self._adaptive_type = value
```
The values associated to AdaptiveType.whatever are already equal to the value passed as input

SMoraisAnsys (approved):
LGTM, I left a minor change proposition for code readability.
Current implementation made me think there was a redirection when it is not the case.

src/pyedb/dotnet/edb_core/edb_data/simulation_configuration.py
SMoraisAnsys: @svandenb-dev Do we commit this changes and merge the PR or are there extra things to be done ?

PR #424, opened 2024-04-29
FIX: edb configuration import stackup

ring630:


svandenb-dev (dismissed):
LGTM

src/pyedb/dotnet/edb_core/materials.py
```diff
@@ -528,7 +528,7 @@ def add_material(self, name: str, **kwargs):
         -------
         :class:`pyedb.dotnet.edb_core.materials.Material`
         """
-        if name in self.__materials:
+        if name.lower() in [i.lower() for i in self.__materials]:
```
SMoraisAnsys, line 531: @ring630 The refactoring of the Material classes removed the need to use `.lower()`. This is because it led with conflict associated to syslib materials and because users could name a material "MyMat" while having to reach it through "mymat". Is there a reason for using `lower()` here ?

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -68,16 +68,161 @@
 logger = logging.getLogger(__name__)
 
 
-class Stackup(object):
+class LayerCollection(object):
+    AUTO_REFRESH = True
```
SMoraisAnsys, line 72: This will create a class parameter and changing the value in one instance will impact other instances. I'm not sure that this is the behavior that you would like to have. If that's the case, please forget the previous comment.

SMoraisAnsys, line 72:
```suggestion
    auto_refresh = True
```
capslock should be used for constant values (here this attribute might be changed by the user)

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -68,16 +68,161 @@
...
+        }
+
+    def update_layout(self, stackup=None):
+        if stackup:
+            self._edb_object = stackup._edb_object
+        self._pedb.layout.layer_collection = self._edb_object
+
+    @pyedb_function_handler
+    def _add_layer(self, add_method, base_layer_name="", **kwargs):
+        layer_clone = kwargs.get("layer_clone")
```
SMoraisAnsys, line 113:
```suggestion
        layer_clone = kwargs.get("layer_clone", None)
```

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -68,16 +68,161 @@
...
+        elif add_method == "add_layer_below":
+            method_above_below = self._edb_object.AddLayerBelow
+        else:  # pragma: no cover
+            return False
+
+        if method_top_bottom:
+            obj = obj if method_top_bottom(obj._edb_object) else False
+        elif method_above_below:
+            obj = obj if method_above_below(obj._edb_object, base_layer_name) else False
+        if self.AUTO_REFRESH:
```
SMoraisAnsys, line 135:
```suggestion
        if self.auto_refresh:
```

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -68,16 +68,161 @@
...
+        # add non stackup layers
+        for _, i in self.non_stackup_layers.items():
+            if i.id == layer_clone.id:
+                lc.AddLayerBottom(layer_clone._edb_object)
+                obj = layer_clone
+            else:
+                lc.AddLayerBottom(i._edb_object)
+
+        self._edb_object = lc
+        if self.AUTO_REFRESH:
```
SMoraisAnsys, line 269:
```suggestion
        if self.auto_refresh:
```

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -68,16 +68,161 @@
...
+        if add_method == "add_layer_top":
+            method_top_bottom = self._edb_object.AddLayerTop
+        elif add_method == "add_layer_bottom":
+            method_top_bottom = self._edb_object.AddLayerBottom
+        elif add_method == "add_layer_above":
+            method_above_below = self._edb_object.AddLayerAbove
+        elif add_method == "add_layer_below":
+            method_above_below = self._edb_object.AddLayerBelow
+        else:  # pragma: no cover
+            return False
```
SMoraisAnsys, line 129:
```suggestion
            logger.error("The way of defining layer addition is not correct")
            return False
```

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -68,16 +68,161 @@
...
+
+        obj = False
+        # Add stackup layers
+        for _, i in self.stackup_layers.items():
+            if i.id == layer_clone.id:  # replace layer
+                add_method(layer_clone._edb_object)
+                obj = layer_clone
+            else:  # keep existing layer
+                add_method(i._edb_object)
+        # add non stackup layers
```
SMoraisAnsys, line 260:
```suggestion
        # Add non stackup layers
```

src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -68,16 +68,161 @@
...
+        for _, i in self.non_stackup_layers.items():
+            if i.id == layer_clone.id:
+                lc.AddLayerBottom(layer_clone._edb_object)
+                obj = layer_clone
+            else:
+                lc.AddLayerBottom(i._edb_object)
+
+        self._edb_object = lc
+        if self.AUTO_REFRESH:
+            self.update_layout()
```
SMoraisAnsys, line 270:
```suggestion
            self.update_layout()
        
        if not obj:
            logger.info("Layer clone was not found in stackup or non stackup layers.")
```

src/pyedb/dotnet/edb_core/stackup.py
ring630: @SMoraisAnsys yes, this is intended. AUTO_REFRESH is a global parameter. With AUTO_REFRESH==False, LayerCollection is saved aside before all layer changes are made.

SMoraisAnsys:
@ring630 @svandenb-dev Please avoid merging this PR until I perform the changes associated to the refactoring of the materials, see #426

SMoraisAnsys (approved):

PR #429, opened 2024-04-29
FIX: stackup import xml

ring630:
@SMoraisAnsys @svandenb-dev this PR is dependent on https://github.com/ansys/pyedb/pull/424.

@svandenb-dev with this implementation. Adding/removing **signal** layers is not allowed.


SMoraisAnsys:
@ring630 What is the status of this PR ? Is it a draft (cf last commit) ? Should we wait for #424 to be merged before reviewing ?

ring630:
> @ring630 What is the status of this PR ? Is it a draft (cf last commit) ? Should we wait for #424 to be merged before reviewing ?

Yes, merge this PR after #424

PR #436, opened 2024-04-29
FIX: get_pins_name_from_net

gkorompi:


src/pyedb/dotnet/edb_core/components.py
```diff
@@ -2402,20 +2402,20 @@ def get_pin_position(self, pin):
...
-    def get_pins_name_from_net(self, pin_list, net_name):
+    def get_pins_name_from_net(self, net_name, pin_list=None):
         """Retrieve pins belonging to a net.
 
         Parameters
         ----------
-        pin_list : list
-            List of pins to check.
+        pin_list : list of EDBPadstackInstance, optional
+            List of pins to check. default is None, in which case all pins are checked
```
SMoraisAnsys, line 2411:
```suggestion
            List of pins to check. The default is ``None``, in which case all pins are checked
```

src/pyedb/dotnet/edb_core/components.py
```diff
@@ -2426,6 +2426,11 @@ def get_pins_name_from_net(self, pin_list, net_name):
 
         """
         pinlist = []
```
SMoraisAnsys, line 2428: It is a bit confusing to see both `pinlist` and `pin_list` in the same method.
Would it make sense to use `pins_name` or `res` instead of `pinlist` ?

src/pyedb/dotnet/edb_core/components.py
SMoraisAnsys, line 2411:
```suggestion
            List of pins to check. The default is ``None``, in which case all pins are checked
```

SMoraisAnsys (changes requested):

ring630 (approved):
LGTM, I approved it. What is the use case of change?

gkorompi:
Generally, it is more usefull to just retrieve all the pins that are connected to a net, rather than checking if a specific pin belongs to this net. Nevertheless, with the new code you can do both. Also, i fixed the format of the returned name, because the component was not mentioned i.e. FCHIP-B11, but only B11 instead. Problem is that B11 also belongs in BGA-B11 and you cannot identify which of the two it references.

PR #439, opened 2024-04-30
FIX: Vias Names

gkorompi:


SMoraisAnsys (approved):
LGTM but it might be worth to extend the scope of this PR to check other tests / add docstrings / ... It feels a bit overkill to have a PR dedicated to modifying "myVia_vullet" into "myVia_bullet" Though it makes totally sense !

Issue #426, opened 2024-04-29
REFACT: Materials to handle GUI interaction


SMoraisAnsys:
Following a discussion with @ring630, it seems that the current implementation of the Material class leads to issues when one wants to use the GUI.

Some major changes of #290 should be reverted to deal with the following extra requirement:
- The GUI does not distinguish between "Copper" and "copper" so **we have to rely on casting material names in lower case on the fly**, otherwise one may end up with errors when working with the GUI;
- The caching mechanism that is used to avoid calling the API is not required as there are very few API calls related to materials.

Note: these requirements were not know when performing #290

SMoraisAnsys:
Closed in #441

PR #441, opened 2024-04-30
REFACTOR: Materials implementation

SMoraisAnsys:
This refactoring of the Materials classes aims at fixing the GUI interactions problems of #426.

Changes:
- Removal of the cache used to work with materials, the resulting approach heavily rely on the EDB API
- Check of material names to avoid ending with two materials with the same name in the GUI.

As a reminder, the GUI is case insensitive and is not able to distinguish two materials like "MatA" and "mata".


ring630 (approved):
LGTM

PR #457, opened 2024-05-07
FEAT: Configuration file nets

svandenb-dev:


ring630 (commented):
@svandenb-dev thanks for adding components and nets. Can you also update _load_components() and _load_nets() in configuration class?

PR #460, opened 2024-05-10
FEAT: Nets config file V2.0

svandenb-dev:


SMoraisAnsys (commented):
If kwargs does not contain nets you'll end up with an errer right ?

svandenb-dev:
> 

Yes

src/pyedb/configuration/cfg_data.py
```diff
@@ -31,17 +35,24 @@ def __init__(self, pedb, **kwargs):
         self.pedb = pedb
         self.edb_comps = self.pedb.components.components
 
-        self.general = None
+        self.general = CfgGeneral(self, kwargs.get("general", None))
         self.boundaries = None
         self.nets = None
+        if kwargs.get("nets"):
+            self.nets = CfgNets(
+                self, kwargs.get("nets", {})["signal_nets"], kwargs.get("nets", {})["power_ground_nets"]
```
SMoraisAnsys, line 43:
```suggestion
                self, kwargs.get("nets", {}).get("signal_nets", []), kwargs.get("nets", {}).get("power_ground_nets", [])
```

src/pyedb/configuration/cfg_pin_groups.py
```diff
@@ -0,0 +1,57 @@
...
+                comp = self._pedb.components.instances[self.reference_designator]
+            else:
+                self._pedb.logger.error("Component not found for creating pin group.")
+                return False
+            pins = [p for p, obj in comp.pins.items() if obj.net_name in self.net]
+            if not self._pedb.siwave.create_pin_group(self.reference_designator, pins, self.name):
+                self._pedb.logger.error(f"Failed to create pin group {self.name}")
+                return False
+            else:
+                self._pedb.logger.error(f"Pin group {self.name} created.")
```
SMoraisAnsys, line 53:
```suggestion
                self._pedb.logger.info(f"Pin group {self.name} created.")
```

SMoraisAnsys (changes requested):
There is still a case to handle with `CfgData` otherwise and error will be triggered.
Btw, is there a reason to call `_apply` instead of `apply` ? You don't seem to leverage inheritance so it might be a good idea to move from one to the other

svandenb-dev:
> There is still a case to handle with `CfgData` otherwise and error will be triggered. Btw, is there a reason to call `_apply` instead of `apply` ? You don't seem to leverage inheritance so it might be a good idea to move from one to the other

The _apply method will remains an internal one called by the main class. I don't want to expose this method to users.

SMoraisAnsys:
> > There is still a case to handle with `CfgData` otherwise and error will be triggered. Btw, is there a reason to call `_apply` instead of `apply` ? You don't seem to leverage inheritance so it might be a good idea to move from one to the other
> 
> The _apply method will remains an internal one called by the main class. I don't want to expose this method to users.

Ok. As a reminder, putting a single `_` does not hide a method/attribute from the user. To enforce that we should add two lower cases `__`. This is something that we should perform in the future.

src/pyedb/configuration/cfg_boundaries.py
```diff
@@ -0,0 +1,115 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from enum import Enum
+
+
+class CfgBoundaries:
+    def __init__(self, pdata, kwargs):
```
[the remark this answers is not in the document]
svandenb-dev: yes done

SMoraisAnsys (dismissed):
LGTM

SMoraisAnsys (approved):
LGTM
@svandenb-dev If possible I'd recommand gathering some of the files into a single one since they are closely related and that would avoid too many tiny files to maintain. Moreover, it could help us with future maintenance. For example, I would probably have though about creating a common interface among Cfg... that follow the same "contract", a.k.a. implementing `apply` method

src/pyedb/configuration/cfg_ports_sources.py
```diff
@@ -30,15 +30,15 @@ def pedb(self):
         return self._pdata.pedb
 
     @pyedb_function_handler
-    def __init__(self, pdata, **kwargs):
```
ring630, line 33: The purpose of **kwargs is to make the input args expandable. Is there a strong reason to change this?

ring630 (commented):
Please also add unittest.

src/pyedb/configuration/cfg_ports_sources.py
svandenb-dev: I got some trouble while implementing. I found it at this harder to read since at the end we know what we are passing. I can try reverting.

ring630: Please revert. Even with dict, we still don't know what is inside.

SMoraisAnsys (approved):
LGTM but still recommending https://github.com/ansys/pyedb/pull/460#pullrequestreview-2055576193

ring630 (changes requested):
There are many errors in unittest, even it is passing. 
image

SMoraisAnsys:
> @svandenb-dev There are many errors in unittest, even it is passing. image

Do you mean that we had test failing but the CICD was not failing ?

ring630:
> > @svandenb-dev There are many errors in unittest, even it is passing. image
> 
> Do you mean that we had test failing but the CICD was not failing ?

Yes, the tested method always return True.

ring630:
> @svandenb-dev There are many errors in unittest, even it is passing. image

@svandenb-dev I fixed this issue.

ring630:
@SMoraisAnsys  it is failing test_18_stackup. Any idea?

>           raise ValueError(f"Material names are case-insensitive and {name.lower()} already exists.")
E           ValueError: Material names are case-insensitive and copper already exists.

ring630 (approved):

PR #462, opened 2024-05-10
FIX: cfg ref size

gkorompi:


svandenb-dev:
@gkorompi did you check this PR ?
https://github.com/ansys/pyedb/pull/456
The component has been refactored please make sure your PR is still valid.

PR #464, opened 2024-05-10
FIX: cfg ref size

gkorompi:


svandenb-dev:
@gkorompi there is a PR pending to be merged with all classes refactoring. Please check when the last one is merged and adjust the code accordingly.

PR #467, opened 2024-05-14
FIX: pin definition

gkorompi:


src/pyedb/dotnet/edb.py
```diff
@@ -4340,3 +4340,28 @@ def definitions(self):
...
+    def change_pin_definition(self, pin):
+        """Change the object definition of the pin from EDBPadstackInstance to PadstackInstance and vice versa.
+        Parameter
+        ----------
+        pin : object EDBPadstackInstance or PadstackInstance
+                Initial definition of Pin.
+        Returns
+        -------
+        pin: secondary definition, depending on the input
+            object EDBPadstackInstance or PadstackInstance.
```
SMoraisAnsys, line 4354:
```suggestion
        """Change the object definition of the pin from EDBPadstackInstance to PadstackInstance and vice versa.

        Parameter
        ----------
        pin : object EDBPadstackInstance or PadstackInstance
                Initial definition of Pin.

        Returns
        -------
        pin: secondary definition, depending on the input
            object EDBPadstackInstance or PadstackInstance.
```

src/pyedb/dotnet/edb.py
```diff
@@ -4340,3 +4340,28 @@ def definitions(self):
...
+        from pyedb.dotnet.edb_core.edb_data.padstacks_data import EDBPadstackInstance
+
+        if isinstance(pin, EDBPadstackInstance):
+            name = pin.name.split("-")
+            for i in self.components.get_pin_from_component(name[0]):
+                if i.GetName() == name[1]:
+                    return i
+        else:
+            name = pin.GetComponent().GetName() + "-" + pin.GetName()
+            for i in [*self.padstacks.pins.values()]:
```
SMoraisAnsys, line 4365:
```suggestion
            for i in self.padstacks.pins.values():
```
Is there a reason to use `*` and then put everything into a list `[ ... ]` ?

SMoraisAnsys (changes requested):

SMoraisAnsys:
@gkorompi I'm not sure to understand how this method does change anything. It feels like it's trying to find something, not changing anything right ?

ring630:
> @gkorompi I'm not sure to understand how this method does change anything. It feels like it's trying to find something, not changing anything right ?

@gkorompi as we discussed.  if get_instances works for you, please cancel this PR.

svandenb-dev (approved):
LGTM

src/pyedb/dotnet/edb_core/components.py
```diff
@@ -2351,7 +2351,11 @@ def get_pin_from_component(self, component, netName=None, pinName=None):
                 and (self.get_aedt_pin_name(p) in pinName or p.GetName() in pinName)
             ]
         else:
-            pins = [p for p in list(component.LayoutObjs) if int(p.GetObjType()) == 1 and p.IsLayoutPin()]
+            pins = [
```
ring630, line 2354: The returned object type is changed. This is not allowed, and it will break customer script. Please revert. Deprecationwarining is added to this method. This method will be removed in the future.

ring630 (changes requested):

Issue #447, opened 2024-05-02
rapter setup returns EDB API objects


ring630:
@svandenb-dev 
https://github.com/ansys/pyedb/blob/bd06c01937129fdf8c042772aa82563c2f1bb8f9/src/pyedb/dotnet/edb_core/edb_data/raptor_x_simulation_setup_data.py#L79

ring630:
@svandenb-dev  can you reuse below property

https://github.com/ansys/pyedb/blob/bd06c01937129fdf8c042772aa82563c2f1bb8f9/src/pyedb/dotnet/edb_core/utilities/simulation_setup.py#L167-L172

svandenb-dev:
Yes will add it when the previous PR is merged

PR #474, opened 2024-05-16
FIX: Raptorx pyedb sweep object property returned

svandenb-dev:
Returned PyEDB frequency sweep object instead of .NET one.

closing #447

### DIFF
--- a/src/pyedb/dotnet/edb_core/utilities/simulation_setup.py
+++ b/src/pyedb/dotnet/edb_core/utilities/simulation_setup.py
@@ -171,7 +171,11 @@ class BaseSimulationSetup(object):
     def frequency_sweeps(self):
         """List of frequency sweeps."""
         temp = {}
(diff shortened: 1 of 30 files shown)


src/pyedb/dotnet/edb_core/utilities/simulation_setup.py
```diff
@@ -171,7 +171,11 @@ def setup_type(self):
     def frequency_sweeps(self):
         """List of frequency sweeps."""
         temp = {}
-        for i in list(self.get_sim_setup_info.SweepDataList):
+        if self.setup_type == "kRaptorX" or "kHFSSPI":
```
SMoraisAnsys, line 174:
```suggestion
        if self.setup_type in ("kRaptorX", "kHFSSPI"):
```

SMoraisAnsys (changes requested):

PR #485, opened 2024-05-20
FIX: rlc create port

gkorompi:


ring630 (changes requested):
Please add unittest to check if rlc is deactived properly

ring630 (approved):

Issue #478, opened 2024-05-17
BUG: modeler.primitive_by_layer crash with None primitives 


svandenb-dev:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

The command is crashing in case the layout returns None primitives. Exception needs to be handled. 

### Steps To Reproduce

NA

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

NA

PR #493, opened 2024-05-21
FIX: primitive by layer crash with none object

svandenb-dev:
Adding bug fix caused by no layer assigned on primitives

closing #478


SMoraisAnsys (approved):
LGTM

PR #498, opened 2024-05-21
REFACTOR: Make plot consistent wrt show

SMoraisAnsys:
Some plot related methods were using a `show` argument and this information was not propagated consistently on the other methods. Changes make it so. Also, the parameter of `plot_matplotlib` associated to the saving of the figure has been updated to be consistent with other plot methods, i.e. from argument `snapshot_path` to `save_plot`.

Note: to make things consistent with previous development, if one provides a path to save the figure. The figure won't be shown, even if argument show is set to True.


svandenb-dev (approved):
LGTM

Issue #352, opened 2024-04-11
Failing to detect correct Ansys version with mesh via plating


svandenb-dev:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

The current EDB version is not working properly to detect Ansys version 2024.1 or higher with setting mesh_via_plating.

### Steps To Reproduce

NA

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

NA

gkorompi:
The information can be transferred to older versions of edb, hence it depends only in the aedt version. If the edb is 23.2 and the aedt version in which imported is 24.2 then the option will be corretly displayed. 
The reason why it does not find the correct version, is that the dataset holds the initial version, and it is updated only when saved. IMO there is no reason to keep the if statement that checks the version, as soon as the edb file correctly transfers the information.

PR #500, opened 2024-05-22
FIX: via mesh

gkorompi:


ring630:
Customer can still run pyedb on older aedt version. Why do you want to remove the version check?

gkorompi:
@ring630 please check comment here: 
https://github.com/ansys/pyedb/issues/352

ring630 (changes requested):
via_mesh_plating should not be allowed in 23.2 and older version

gkorompi:
It will not crash, it depends solely on the AEDT version that is used to open the design. If you define the variable in 23.2 edb and you launch it with 24.2 aedt, then the option will be there. Also, the dataset is not updated until saved, so it makes no sense to check it as soon as it is not updated! @svandenb-dev  what's your opinion?

src/pyedb/dotnet/edb_core/edb_data/hfss_simulation_setup_data.py
```diff
@@ -977,19 +977,14 @@ def via_mesh_plating(self):
         -------
         bool
         """
-        if self._parent._pedb.version[0] >= 10:
-            return self._via_settings.ViaMeshPlating
-        else:
-            self._parent._pedb.logger.error("Property only supported on Ansys release 2024R1 and later")
-            return False
+        return self._via_settings.ViaMeshPlating
+        self._parent._pedb.logger.warning("Property only supported on Ansys release 2024R1 and later")
```
SMoraisAnsys, line 981: Just so you know, this log warning will never be triggered as a return if performed just before.

ring630:
@gkorompi I am not convinced by this PR. I closed it for now.

ring630:
@gkorompi #515

PR #501, opened 2024-05-22
REFACTOR: Component configuration file

svandenb-dev:
Configuration file


ring630 (approved):
LGTM. Thanks for remove _load_sources(). Would you consider merge the small py files as suggested by Sebastien?

PR #505, opened 2024-05-23
FEATURE: Extend create port with name

gkorompi:


svandenb-dev (changes requested):
You are missing the test. Just add a simple test in the existing one I guess. Otherwise with doing that even small things we will decrease the codecov

SMoraisAnsys (approved):

svandenb-dev (approved):
LGTM

PR #511, opened 2024-05-24
FIX: bug in uniitest

ring630:


SMoraisAnsys (approved):
LGTM, thanks for catching this error in our test base !

PR #527, opened 2024-05-30
FIX: stackup load

ring630:
The cause is there is a weird behavior in EDB API. It seems it doesn't allow the existence of two layer collections.


svandenb-dev (approved):
LGTM

SMoraisAnsys (approved):

PR #537, opened 2024-06-04
FIX: gds fix

gkorompi:


src/pyedb/dotnet/edb.py
```diff
@@ -1523,12 +1523,8 @@ def import_gds_file(
             ``True`` when successful, ``False`` when failed.
 
         """
-        if tech_file or map_file:
-            control_file_temp = os.path.join(tempfile.gettempdir(), os.path.split(inputGDS)[-1][:-3] + "xml")
-            ControlFile(xml_input=control_file, tecnhology=tech_file, layer_map=map_file).write_xml(control_file_temp)
-        elif tech_file:
-            self.logger.error("Technology files are supported only in Linux. Use control file instead.")
```
SMoraisAnsys, line 1530: Can technology file be used with linux now ?

gkorompi: I dont think so, but why to return False if tech_file is used? Maybe we should add it as a warning?

SMoraisAnsys: Then we should probably refactor to leverage `is_linux` and trigger the error message if `tech_file and not is_linux`. Otherwise, users won't know that technology files are only supported in linux.
If not then we can proceed with your proposed changes.

SMoraisAnsys (changes requested):

src/pyedb/dotnet/edb.py
```diff
@@ -1527,12 +1527,11 @@ def import_gds_file(
             ``True`` when successful, ``False`` when failed.
 
         """
-        if tech_file or map_file:
-            control_file_temp = os.path.join(tempfile.gettempdir(), os.path.split(inputGDS)[-1][:-3] + "xml")
-            ControlFile(xml_input=control_file, tecnhology=tech_file, layer_map=map_file).write_xml(control_file_temp)
-        elif tech_file:
+        if is_linux and tech_file:
```
SMoraisAnsys, line 1530:
```suggestion
        if not is_linux and tech_file:
```
It's the opposite, we only want to fail on non linux OS.

SMoraisAnsys (changes requested):

SMoraisAnsys (approved):

PR #544, opened 2024-06-06
FEAT: Add deprecate decorator

### LINKED ISSUE #540: REFACTOR: Remove pyedb function handler
Contrary to `pyaedt` where there seems to be value to keep a function handler, using a function handler in `pyedb` does not seem to add much value (for example, no value to redirect info to UI or to be more fault tolerant). More over, it hides under the hood issues. That's something that we found out last week when working on an example over Materials.

Change proposed: removal of the function handler and direct handling of the raised methods.

SMoraisAnsys:
This PR si a first step to remove the `pyedb_function_handler`. At some point we wanted to leverage the function handler to provide argument deprecation capability (like in `pyaedt`) but this PR should handle this aspect.

Partially adress #540


ring630 (approved):
LGTM

PR #545, opened 2024-06-07
CI: Use Trusted Publisher to release package

SMoraisAnsys:
We are planning to transition from using secret tokens to utilizing Trusted Publisher for our public package distribution.

This change will enhance:

- our security as secrets can be exposed inadvertently through leaks;
- our workflows as managing and rotating secrets requires additional effort (and can be error-prone).

Once this is merged, I'll contact the right person to proceed with updating our `pypi` settings.


svandenb-dev (approved):
LGTM

PR #549, opened 2024-06-07
FIX: unittest arbitrary wave ports

ring630:


ring630:
@svandenb-dev  @SMoraisAnsys I fixed the bug in unittest. I don't understand how it did pass.

SMoraisAnsys (approved):
LGTM.
I assume that it wasn't failing because the test was simply not executed. `pytest` expects tests to start with `test_`, since it didn't, the function was probably never called....

PR #556, opened 2024-06-10
FIX: Failure in CICD with plot on matplotlib

### LINKED ISSUE #516: TESTS: There seem to be something wrong with test_padstaks_plot_on_matplotlib
From what I've been able to see with recent PR merge, the CICD passed the pipeline using pyedb-ci-0 but when the PR gets squashed and merged with a single commit, in the main branch, the CICD failed the pipeline when using pyedb-ci-1.

We need to look into what's wrong with this VM to avoid such false positive behavior (be it from pyedb-ci-0 or 1) !

SMoraisAnsys:
Seems like some recent changes related to `plot` had a side effect on the test behavior.
This PR ensures that every job in CICD has environment variable `PYEDB_CI_NO_DISPLAY` set to `"1"`. This way we ensure that we use `Agg` backend and shouldn't come across this error again.

Closes #516

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -77,7 +77,7 @@ jobs:
 
   legacy-tests:
     name: Test dotnet
-    runs-on: [ Windows, self-hosted, pyedb, pyedb-ci-1 ]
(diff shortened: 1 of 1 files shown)


svandenb-dev (approved):
LGTM

Issue #480, opened 2024-05-17
Improve import material


ring630:
### Description of the current limitations

test_18_stackup is producing errors, but unittest is still passing. The cause of failure is material name is sensitive. Failed to set COPPER since copper is already in material library. 

My suggestion is to update material properties when material already exists.

@SMoraisAnsys  @svandenb-dev

SMoraisAnsys:
We can discuss about this but I would rather avoid that. If I'm not mistaken, there is a way to remove a material and I feel like it would make more sense to let the user remove something to create another one. This way, we would ensure that he is aware of the case sensitive issue.

@ring630 Can we discuss about this when I'm back from PyCon US? I probably don't have (again) enough knowledge on the materials aspects and your point of view could help me understand how we should do things.

ring630:
@SMoraisAnsys Material is a special case, and it was not handled properly. I am glad you are fixing it. Let's discuss further. 

Here I list the facts.
1, AEDT UI is not case sensitive. COPPER and copper cannot coexist.
2, EDB is case senstitive. It should be prevented to have both COPPER and copper. 
3, extra care is required to rename COPPER to copper if COPPER already exist. Layers are using COPPER.

SMoraisAnsys:
Summaryze changes required:
- allow changing the name of a material
- allow a configuration to "update" the material information (either rename + update data or delete + create)
- use lower case when loading AMAT file

SMoraisAnsys:
Reopening this issue as I don't think it is fully handled.

svandenb-dev:
what is the status @SMoraisAnsys ?

PR #557, opened 2024-06-10
FIX: unittest materials

ring630:


SMoraisAnsys (approved):
This will help users for sure !

For cross-reference purposes, PR is related to #480

PR #562, opened 2024-06-10
DOCS: Fix path to resoures figures

SMoraisAnsys:
Some path were using `Resources` in the path when the directory's name is `resources`.
This PR should fix the failing paths.


svandenb-dev (approved):
LGTM

PR #563, opened 2024-06-10
REFACTOR!: Remove function handler

### LINKED ISSUE #540: REFACTOR: Remove pyedb function handler

SMoraisAnsys:
Closes #540, #574 and #575


SMoraisAnsys:
The scope of this PR has been extended to fix issues found while removing the function handler:
- @svandenb-dev There is an issue when loading IPC model (see #574), commit 097c7156d951946e5bb01fb7e84cc6db8eca88a4 should fixe the issue.
- @ring630 There is an issue when creating a package ref from a component (see #575), commit 6277f3ce0 should fix the issue.

Let me know if we need extra changes !

svandenb-dev (dismissed):
LGTM

PR #565, opened 2024-06-10
REFACTOR: Pyedb hierarchy terminal refactoring

svandenb-dev:
First draft


src/pyedb/dotnet/edb_core/terminal/bundle_terminal.py
```diff
@@ -0,0 +1,51 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from pyedb.dotnet.edb_core.terminal.edge_terminal import EdgeTerminal
+
+
+class BundleTerminal(object):
```
SMoraisAnsys, line 26:
```suggestion
from pyedb.dotnet.edb_core.terminal.edge_terminal import EdgeTerminal
from pyedb.dotnet.edb_core.terminal.terminal import Terminal

class BundleTerminal(Terminal):
```

Shouldn't this class (and the following) inherit from `Terminal` ?
Seems to be the case in the EDB API.

src/pyedb/dotnet/edb_core/terminal/edge_terminal.py
```diff
@@ -0,0 +1,49 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from pyedb.dotnet.edb_core.general import convert_py_list_to_net_list
+
+
+class EdgeTerminal(object):
```
SMoraisAnsys, line 26: `Terminal` ?

src/pyedb/dotnet/edb_core/terminal/padstack_instance_terminal.py
```diff
@@ -0,0 +1,87 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from pyedb.dotnet.edb_core.edb_data.padstacks_data import EDBPadstackInstance
+from pyedb.generic.general_methods import generate_unique_name
+
+
+class PadstackInstanceTerminal(object):
```
SMoraisAnsys, line 27: `Terminal` ?

src/pyedb/dotnet/edb_core/terminal/pingroup_terminal.py
```diff
@@ -0,0 +1,57 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+
+class PinGroupTerminal(object):
```
SMoraisAnsys, line 24: `Terminal` ?

src/pyedb/dotnet/edb_core/terminal/point_terminal.py
```diff
@@ -0,0 +1,85 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+
+class PointTerminal(object):
```
SMoraisAnsys, line 24: `Terminal` ?

SMoraisAnsys (dismissed):
Unless I'm wrong, there is missing inheritance.
Otherwise, you might want to get rid of `super` calls since your classes don't inherit from something.

ring630 (commented):
@svandenb-dev , did you just move the files or changed anything in the class?

src/pyedb/dotnet/edb_core/terminal/bundle_terminal.py
ring630: That's a good catch.

svandenb-dev: I think we are creating circular import ?

svandenb-dev:
> @svandenb-dev , did you just move the files or changed anything in the class?

Just moved

SMoraisAnsys (approved):
LGTM
Some property getter are returning a specific value if a condition is valid (leading to returning `None` otherwise), see for example `Terminal` class and `layer` property. Could we log a warning in those cases ?

svandenb-dev:
> LGTM Some property getter are returning a specific value if a condition is valid (leading to returning `None` otherwise), see for example `Terminal` class and `layer` property. Could we log a warning in those cases ?

Thanks @SMoraisAnsys I have added some wrnings

SMoraisAnsys:
> > LGTM Some property getter are returning a specific value if a condition is valid (leading to returning `None` otherwise), see for example `Terminal` class and `layer` property. Could we log a warning in those cases ?
> 
> Thanks @SMoraisAnsys I have added some wrnings

Added changes consist in returning `False`, not logging warnings.
It'll "work" though.

SMoraisAnsys (approved):
LGTM. Actual logs would be great but returning `False` also works

PR #566, opened 2024-06-11
FEAT: Added find_points_along_lines method in GeometryOperators class

Alberto-DM:


svandenb-dev (dismissed):
LGTM

svandenb-dev (dismissed):
LGTM

SMoraisAnsys (dismissed):
Following @svandenb-dev approval, I'll remove my comment.

The main point of my remark was base on two things:

- the fact that we have other test mocking simple data (and I proposed to follow that approach)
- the fixture that you use might is a bit complex

However, I don't want to nitpick and your code and test are clear and well implemented. Thanks for the contribution !

Alberto-DM:
> Following @svandenb-dev approval, I'll remove my comment.
> 
> The main point of my remark was base on two things:
> 
> * the fact that we have other test mocking simple data (and I proposed to follow that approach)
> * the fixture that you use might is a bit complex
> 
> However, I don't want to nitpick and your code and test are clear and well implemented. Thanks for the contribution !

I simplified the fixture. So the second point is addressed.

svandenb-dev (approved):
LGTM

PR #588, opened 2024-06-18
FIX: add mesh operation

ring630:


src/pyedb/dotnet/edb_core/utilities/simulation_setup.py
```diff
@@ -140,11 +140,6 @@ def _set_edb_setup_info(self, edb_setup_info):
     def _update_setup(self):
         """Update setup in EDB."""
         # Update mesh operation
-        if self._setup_type == "kHFSS":
```
SMoraisAnsys, line 143: Just a question, why do you remove this lines of code ?

SMoraisAnsys (approved):

src/pyedb/dotnet/edb_core/utilities/simulation_setup.py
ring630: @SMoraisAnsys I removed the caching self._mesh_operations. These lines are no longer needed.

ring630: @SMoraisAnsys it is a mistake. I will add them back.

PR #589, opened 2024-06-20
DOCS: layout validation documentation exposed

svandenb-dev:
layout validation documentation exposed


SMoraisAnsys (approved):
LGTM but we might want to make a big PR to add all the missing documentation.

PR #593, opened 2024-06-20
FIX: Fixes 2 issues in ipc2581 exports.

maxcapodi78:
Tons of warning on layers not present in stackup and the failure to add_profile because Outline is a non_stackup_layer


svandenb-dev (approved):
LGTM

PR #594, opened 2024-06-21
FIX: numpy version 2.0 issue fixed

svandenb-dev:


Samuelopez-ansys (approved):

SMoraisAnsys (approved):
LGTM but could you extend the issue description so that we can know what is going wrong ?
AFAIK the tests passed correctly with version `numpy>=2.0` so it would be great to know what broke.

PR #605, opened 2024-06-25
FEAT: added expand method to class PrimitivesDotNet

maxcapodi78:
I've also added an overload method for path and circles which changes respectively the width and the radius with the offset


svandenb-dev (approved):
LGTM

PR #607, opened 2024-06-25
REFACTOR: Primitive scale

svandenb-dev:
Moving scale to EDBPrimitive to be consistent with expand method.


maxcapodi78 (approved):

src/pyedb/dotnet/edb_core/dotnet/primitive.py
```diff
@@ -376,6 +376,39 @@ def expand(self, offset=0.001, tolerance=1e-12, round_corners=True, maximum_corn
...
+        -------
+        bool
+           ``True`` when successful, ``False`` when failed.
+        """
+        if not isinstance(factor, str):
+            factor = float(factor)
+            polygon_data = self.polygon_data.create_from_arcs(self.polygon_data.edb_api.GetArcData(), True)
+            if not center:
+                center = self.polygon_data.edb_api.GetBoundingCircleCenter()
+                if center:
```
SMoraisAnsys, line 399: Can you log an error when this if statement is not valid ? Or assume that it should work (I mean pass the if statement) and remove this condition ?

SMoraisAnsys (changes requested):
Why are you adding an empty file to the PR ?
I left a comment on the scale function.

src/pyedb/dotnet/edb_core/dotnet/primitive.py
svandenb-dev: Done

ring630 (approved):
LGTM

SMoraisAnsys (commented):
@svandenb-dev Please remove the empty file as asked in my previous review

ring630 (commented):
Please fix the empty file.

ring630 (approved):

PR #610, opened 2024-06-25
FEAT: export bom

ring630:
Changed default behavior of component.value.
- Old. Return a default value when RLC component doesn't have a value. This introduces confusion.
- New. Return None when RLC component doesn't have a value.

### DIFF
--- a/src/pyedb/configuration/cfg_components.py
+++ b/src/pyedb/configuration/cfg_components.py
@@ -51,12 +51,13 @@ class CfgRlcModel(CfgBase):
 
 
 class CfgComponent(CfgBase):
-    protected_attributes = ["reference_designator"]
+    protected_attributes = ["reference_designator", "definition", "location", "angle", "placement_layer"]
 
     def __init__(self, **kwargs):
         self.enabled = kwargs.get("enabled", None)
 
         self.reference_designator = kwargs.get("reference_designator", None)
+        self.definition = kwargs.get("definition", None)
         self.type = kwargs.get("part_type", None)
         self.value = kwargs.get("value", None)
         self.port_properties = CfgPortProperties(**kwargs["port_properties"]) if "port_properties" in kwargs else None
@@ -66,6 +67,23 @@ class CfgComponent(CfgBase):
         rlc_models = kwargs.get("rlc_model", [])
 
         self.rlc_model = [CfgRlcModel(**rlc_m) for rlc_m in rlc_models]
+
+        self.x_location, self.y_location = kwargs.get("location", [None, None])
+        self.angle = kwargs.get("angle", None)
+        self.placement_layer = kwargs.get("placement_layer", None)
+
+    def export(self):
+        data_comp = {}
+        data_comp["enabled"] = self.enabled
+        data_comp["reference_designator"] = self.reference_designator
+        data_comp["definition"] = self.definition
+        data_comp["type"] = self.type
+        data_comp["value"] = self.value
+        data_comp["x_location"] = self.x_location
+        data_comp["y_location"] = self.y_location
+        #data_comp["angle"] = self.angle
+        data_comp["placement_layer"] = self.placement_layer
+        return data_comp
 
 
 class CfgComponents:
@@ -127,3 +145,23 @@ class CfgComponents:
                         setattr(c_db, attr, value)
                     else:
                         raise AttributeError(f"'{attr}' is not valid component attribute.")
+
+    def _load_data_from_db(self):
+        self.components = []
+        comps_in_db = self._pedb.components
+        for _, comp in comps_in_db.components.items():
+            cfg_comp = CfgComponent(enabled=comp.enabled,
+                                    reference_designator=comp.name,
+                                    part_type=comp.type,
+                                    value=comp.value,
+                                    definition=comp.component_def,
+                                    location=comp.location,
+                                    placement_layer=comp.placement_layer)
+            self.components.append(cfg_comp)
+
+    def get_data_from_db(self):
+        self._load_data_from_db()
+        data = []
+        for comp in self.components:
+            data.append(comp.export())
+        return data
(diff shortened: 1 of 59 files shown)


src/pyedb/configuration/cfg_components.py
```diff
@@ -51,12 +51,13 @@ def __init__(self, **kwargs):
 
 
 class CfgComponent(CfgBase):
-    protected_attributes = ["reference_designator"]
+    protected_attributes = ["reference_designator", "definition", "location", "angle", "placement_layer"]
```
SMoraisAnsys, line 54: After some though I'm wondering if we shouldn't drop the `protected_attributes` approach.
It's a bit confusing to ask for attributes and not retrieving the whole attributes but a subset of them without knowing that there is such filter (from the method name).

Since it's mainly associated to the current `get_attributes` and `set_attributes`, why not reworking this to use:
- `to_dict` (equivalent to `get_attributes`) but without exclude.
- `update` (equivalent to `set_attributes`) where we would have a call to `to_dict` with inner filtering as we don't want to update certain attributes.

That way the `protected_attributes` won't be a class attribute that is accessible by the user. The attributes we don't want them to update would be set in stone inside `update` and we could trigger warnings when one wants to do it.
What do you think ? If you agree, I can help with that.

src/pyedb/configuration/cfg_components.py
```diff
@@ -67,6 +68,23 @@ def __init__(self, **kwargs):
 
         self.rlc_model = [CfgRlcModel(**rlc_m) for rlc_m in rlc_models]
 
+        self.x_location, self.y_location = kwargs.get("location", [None, None])
+        self.angle = kwargs.get("angle", None)
+        self.placement_layer = kwargs.get("placement_layer", None)
+
+    def export(self):
```
SMoraisAnsys, line 75: What is the scope of this `export` method ? Can you rename it or add docstring such that we know why it doesn't contain all attributes data ?

src/pyedb/dotnet/edb_core/cell/hierarchy/component.py
```diff
@@ -56,9 +57,9 @@ class EDBComponent(object):
 
     """
 
-    def __init__(self, pedb, cmp):
-        self._pedb = pedb
-        self.edbcomponent = cmp
+    def __init__(self, pedb, edb_object):
```
SMoraisAnsys, line 60: Just to know, what's the difference here between `pedb` and `edb_object` ? The first is the `Edb` instance while the second is any object from the API ?

src/pyedb/dotnet/edb_core/cell/hierarchy/hierarchy_obj.py
```diff
@@ -0,0 +1,45 @@
...
+    def component_def(self):
+        """Component definition."""
+        return self._edb_object.GetComponentDef().GetName()
+
+    @property
+    def location(self):
+        """XY Coordinates."""
+        flag, x, y = self._edb_object.GetLocation()
+        if flag:
+            return [x, y]
```
SMoraisAnsys, line 40: Can you log an error if flag is not working in this condition ?

src/pyedb/dotnet/edb_core/cell/hierarchy/component.py
```diff
@@ -395,12 +399,7 @@ def value(self):
         """
         if self.model_type == "RLC":
             if not self._pin_pairs:
-                if self.type == "Inductor":
-                    return 1e-9
-                elif self.type == "Resistor":
-                    return 1e6
-                else:
-                    return 1
+                return
```
SMoraisAnsys, line 402: Can you justify (in the PR) this change so that we know why previous default behavior is changed ?

SMoraisAnsys (changes requested):
LGTM but I would like to discuss about the `protected_attributes` (maybe in another issue / PR though).
Can you add extra tests on `HierarchObj` ?

src/pyedb/configuration/cfg_components.py
ring630: @SMoraisAnsys  Great idea. I will change according to your suggestion in a separate PR.

src/pyedb/dotnet/edb_core/cell/hierarchy/component.py
ring630: pedb is pyedb.edb.Edb

edb_object is EDB API object. In this case, it is   Ansys.Ansoft.Edb.Cell.Hierarchy.Component.

src/pyedb/configuration/cfg_components.py
SMoraisAnsys: > @SMoraisAnsys Great idea. I will change according to your suggestion in a separate PR.

I was supposed to work on that at some point (if I'm not wrong) while leveraing `pydantic`.
Mind if I do it as planned ? Sorry for the delay in handling that though !

src/pyedb/dotnet/edb_core/cell/hierarchy/hierarchy_obj.py
```diff
@@ -0,0 +1,48 @@
...
+        """Component definition."""
+        return self._edb_object.GetComponentDef().GetName()
+
+    @property
+    def location(self):
+        """XY Coordinates."""
+        flag, x, y = self._edb_object.GetLocation()
+        logging.warning(f"Failed to get location of '{self.name}'.")
+        if flag:
+            return [x, y]
```
SMoraisAnsys, line 43:
```suggestion
        if flag:
            return [x, y]
        else:
            logging.error(f"Failed to get location of '{self.name}'.")
```

SMoraisAnsys (changes requested):

src/pyedb/dotnet/edb_core/cell/hierarchy/hierarchy_obj.py
SMoraisAnsys: @ring630 The log is done too early, we need to trigger it iff flag was not working (I assume)

PR #632, opened 2024-07-02
FEAT: edb configuration export ports and sources

ring630:


svandenb-dev (changes requested):
All is good great work. See my comment on the method name only.

svandenb-dev (approved):
LGTM

PR #649, opened 2024-07-08
REFACTOR: layout 

ring630:
-  Separated class Layout from class EDBLayout.
- Moved/refactored methods from class dotnet.layout.LayoutDotNet into class Layout
- Deleted redundant file dotnet.layout.py
- rename EDBLayout to Moderler. 
- changed - one py file only has one class.   
  - This is due to the fact that there are many levels of inheritance, which leads to single file too big. Small classes of cause can be put in one file.
  - Circular import can be avoided if import is done appropriately.
- Moved BuildSimulationConfiguration unittest into a separate file. 
- Removed many redundant properties. These properties should be inherit from parent class.

Properties of Layout returns pyedb classes. These properties should be reused as much as possible. There are methods in class Modeler are dealing with EDB API classes. These methods should be refactored in a separate PR.


svandenb-dev (approved):
LGTM

PR #652, opened 2024-07-10
FEAT: Added getitem to primitives. fixed bug in getitem for padstack. 

maxcapodi78:
Added cache data of instances and definitions properties to make it faster to retrieve info


src/pyedb/dotnet/edb_core/layout.py
```diff
@@ -124,11 +147,13 @@ def primitives(self):
         list of :class:`pyedb.dotnet.edb_core.edb_data.primitives_data.EDBPrimitives`
             List of primitives.
         """
-        _prims = []
+        if len(self._primitives) == len(self._layout.primitives):
+            return self._primitives
```
SMoraisAnsys, line 151: Is it possible that we end up adding an element to `self._layout.primitives` and the removing another one, making `self._layout.primitives` having the same length but with a different content ? If yes, then I think we should leverage this if condition as one might end up with incorrect cached data.

src/pyedb/dotnet/edb_core/padstack.py
```diff
@@ -183,12 +185,14 @@ def definitions(self):
             List of definitions via padstack definitions.
 
         """
-        _padstacks = {}
+        if len(self._definitions) == len(self._pedb.padstack_defs):
```
SMoraisAnsys, line 188: Same question as previous comment

src/pyedb/dotnet/edb_core/padstack.py
```diff
@@ -217,11 +221,13 @@ def instances(self):
 
         """
 
-        padstack_instances = {}
         edb_padstack_inst_list = self._pedb.layout.padstack_instances
+        if len(self._instances) == len(edb_padstack_inst_list):
```
SMoraisAnsys, line 225: Same question as first comment

src/pyedb/dotnet/edb_core/layout.py
maxcapodi78: I've added a cleanup of the variables on the delete operation so we should be safe enough.

ring630 (changes requested):
@maxcapodi78 there is a method FindById which gets primitives from layout fast. getitem can benefit from it and gets much faster. It is implemented in another PR https://github.com/ansys/pyedb/pull/649. That PR is a complete refactor of EDBLayout. Please wait until it's merged.

svandenb-dev (approved):
LGTM

src/pyedb/dotnet/edb_core/cell/layout_obj.py
```diff
@@ -89,6 +89,9 @@ def id(self):
     def delete(self):
         """Delete this primitive."""
         self._edb_object.Delete()
+        self._pedb.modeler._primitives = []
+        self._pedb.padstacks._instances = []
+        self._pedb.padstacks._definitions = []
```
SMoraisAnsys, line 94: Should we also do that when on updates `_edb_properties` since the value might be changed ? I'm wondering if this could lead to a cached object that does not contain the correct value this time.

maxcapodi78: everything in the object is not cached. So there is no need to refresh when you edit the property

ring630 (approved):

PR #661, opened 2024-07-11
REFACTOR: removed is_ironpython and inside_desktop variables as it is not anymore

maxcapodi78:


SMoraisAnsys (commented):
@maxcapodi78 Don't we have to keep compatibility for the time being as `pyaedt` is still IronPython compatible ? This should be changed pretty soon though.

svandenb-dev (approved):
LGTM

maxcapodi78:
> @maxcapodi78 Don't we have to keep compatibility for the time being as `pyaedt` is still IronPython compatible ? This should be changed pretty soon though.

@SMoraisAnsys no because Ironpython in AEDT doesn't have pip support, so it's already excluded from actual support

ring630 (approved):

PR #675, opened 2024-07-15
FEAT: Improved auto_parametrize_design method.

maxcapodi78:
Variable renaming and also the trace_net_filters now applies also to padstack definitions.

### DIFF
--- a/src/pyedb/dotnet/edb.py
+++ b/src/pyedb/dotnet/edb.py
@@ -4150,6 +4150,7 @@ class Edb(Database):
         if isinstance(trace_net_filter, str):
             trace_net_filter = [trace_net_filter]
         parameters = []
+
         def _apply_variable(orig_name, orig_value):
             if use_relative_variables:
                 var = f"{orig_name}_delta"
@@ -4164,7 +4165,7 @@ class Edb(Database):
             if use_relative_variables:
                 return f"{orig_value}+{var}", var
             else:
-                 return var, var
+                return var, var
 
         if layers:
             if not layer_filter:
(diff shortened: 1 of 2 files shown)


src/pyedb/dotnet/edb.py
```diff
@@ -4122,20 +4123,25 @@ def auto_parametrize_design(
...
         List(str)
             List of all parameters name created.
         """
+        if isinstance(trace_net_filter, str):
+            trace_net_filter = [trace_net_filter]
         parameters = []
         if layers:
             if not layer_filter:
-                _layers = self.stackup.stackup_layers
+                _layers = self.stackup.layers
```
svandenb-dev, line 4140: Why are we taking also non stackup layers ? I am nit sure we can apply parameters on them

src/pyedb/dotnet/edb.py
```diff
@@ -4122,20 +4123,25 @@ def auto_parametrize_design(
...
             else:
                 if isinstance(layer_filter, str):
                     layer_filter = [layer_filter]
-                _layers = {k: v for k, v in self.stackup.stackup_layers.items() if k in layer_filter}
+                _layers = {k: v for k, v in self.stackup.layers.items() if k in layer_filter}
```
svandenb-dev, line 4144: same question

svandenb-dev (approved):
LGTM

src/pyedb/dotnet/edb.py
maxcapodi78: stackup_layers is deprecated
it returns layers

src/pyedb/dotnet/edb.py
maxcapodi78: image

svandenb-dev (approved):
LGTM

svandenb-dev (approved):
LGTM

PR #683, opened 2024-07-18
FIX: Change line 815 and 816.  Update the way 3D object is created

amichel0205:
Equation to lowest the circle is applied to rad 1 to respect shape of microvias from litterature.
With the current code the via was drawn like this

closing #682


svandenb-dev (approved):
LGTM

PR #688, opened 2024-07-24
FIX: Material with import stackup file

svandenb-dev:
closing #680


src/pyedb/dotnet/edb_core/edb_data/simulation_configuration.py
```diff
@@ -2478,7 +2477,7 @@ def _read_cfg(self):  # pragma: no cover
                 cfg_lines = cfg_file.read().split("\n")
                 for line in cfg_lines:
                     if line.strip() != "":
-                        if line.find("="):
+                        if not line.find("=") == -1:
```
maxcapodi78, line 2480: I would use different approach
```suggestion
                        if line.find("=")>0:
```

maxcapodi78 (approved):

PR #692, opened 2024-07-24
FEAT: selected_angles parameter for find_points_along_lines method

Alberto-DM:
Added selected_angles argument to find_points_along_lines method.
User can decide the orientation of the lines that are detected by the method.


svandenb-dev (approved):
LGTM

PR #702, opened 2024-07-30
FEAT: added offset of vias in auto_parametrize_design

maxcapodi78:


SMoraisAnsys (changes requested):
LGTM
Can you add an example to test this new feature ?

maxcapodi78:
> LGTM Can you add an example to test this new feature ?

@SMoraisAnsys done it. I didn't do because the default is true but it makes sense to test both cases

PR #707, opened 2024-08-02
REFACTOR: edb configuration documentation

ring630:
- group edb configuration examples into two
- format fix in two examples


SMoraisAnsys:
Updating the branch after main fix

SMoraisAnsys (changes requested):
It might be better to set the thumbnail directly from the conf.py file. You just have to update the _static folder with the png. This will simplify the rst files to make them more readable.

Here is an example:

```python
# Sphinx gallery customization
nbsphinx_thumbnails = {
    "examples/use_configuration/import_ports": "_static/thumbnails/101_getting_started.png",
}
```

ring630:
> ```python
> thumbnails
> ```

@SMoraisAnsys  Done

SMoraisAnsys (changes requested):
I downloaded the artifact and the result is not working as expected, see bellow

image

Can you try it locally to fix it ?

SMoraisAnsys:
> > ```python
> > thumbnails
> > ```
> 
> @SMoraisAnsys Done

You didn't modify the conf.py file, see the result in my previous comment.

ring630:
> > > ```python
> > > thumbnails
> > > ```
> > 
> > 
> > @SMoraisAnsys Done
> 
> You didn't modify the conf.py file, see the result in my previous comment.

Ok. Updated.

SMoraisAnsys:
Did you test it locally ? I downloaded the artifact and got the following result

image

Is it from my side or something you also experience ?

ring630:
@SMoraisAnsys I see the same on my local machine. Not sure what I should do to fix it.

SMoraisAnsys:
@ring630 I tested locally and the fix of path worked !

PR #720, opened 2024-08-09
CI: Revert upload artifact to v4 (unpin)

SMoraisAnsys:
Recently we had to pin the version of the upload artifact, this is no longer required.


ring630 (approved):
LGTM

PR #721, opened 2024-08-09
REFACTOR: primitives 

ring630:
- Replaced legacy class PrimitiveDotNet with class Primitive
- Replaced legacy class PathDoNet with class Path
- Removed try-except to avoid hidden code error
- Avoid EDB API objects as input arguments


ring630:
@SMoraisAnsys @svandenb-dev this pr is ready for review now : )

svandenb-dev (approved):
LGTM

PR #713, opened 2024-08-06
CHORE: Update code base to use AEDT 24.2

SMoraisAnsys:
Upgrading our code base to use AEDT 24.2 as default version.

@svandenb-dev @ring630 One of the changes that was performed comes from the fix of @Samuelopez-ansys in [68b393a](https://github.com/ansys/pyedb/pull/713/commits/68b393a08d45f89ad3a843cad939ce52b1ac3eee) It seems that AEDT 24.2 has changed the way it works with waveport naming. Previously, it changed the name on the fly and it is not the case anymore with 24.2. The fix consists in mocking what was performed previously with the naming so that 24.1 and 24.2 are both working !

PR #724, opened 2024-08-16
FIX: create differential pair

ring630:


SMoraisAnsys (commented):
@ring630 Do you mind testing by adding this commit to https://github.com/ansys/pyedb/pull/713 ? The current changes are only tested for 24.1 and I'm not sure we won't need to update the behavior of other things when going to 24.2.

ring630:
> @ring630 Do you mind testing by adding this commit to #713 ? The current changes are only tested for 24.1 and I'm not sure we won't need to update the behavior of other things when going to 24.2.

@SMoraisAnsys good idea. I merged the changes into #713

Issue #727, opened 2024-08-19
FEAT: Evaluate ESL and ESR from component libraries


svandenb-dev:
### Description of the feature

Evaluate ESL and ESR from component libraries

PR #730, opened 2024-08-20
FEAT: Ansys components libraries ESL ESR computation

svandenb-dev:
Ansys components libraries ESL ESR computation

closing #727


src/pyedb/component_libraries/ansys_components.py
```diff
@@ -33,16 +33,61 @@ def __init__(self, name, index, sbin_file):
...
+        """Returns equivalent serial resistor for capacitor only."""
+        if self.type == "Capacitor":
+            z11 = 1 / self.s_parameters.y[:, 0, 0]
+            return np.abs(abs(np.abs(np.real((z11.min()))) - 50))
+        else:
+            return 0.0
+
+    @property
+    def f0(self):
+        """Returns capacitor self resonant frequency in Hz."""
```
SMoraisAnsys, line 58:
```suggestion
        """Return a skrf.network.Network object.
        
        See `scikit-rf documentation <https://scikit-rf.readthedocs.io/en/latest/api/network.html#network-class>`_.
        """
        if not self._s_parameters:
            self._extract_impedance()
        return self._s_parameters

    @property
    def esr(self):
        """Return the equivalent serial resistor for capacitor only."""
        if self.type == "Capacitor":
            z11 = 1 / self.s_parameters.y[:, 0, 0]
            return np.abs(abs(np.abs(np.real((z11.min()))) - 50))
        else:
            return 0.0

    @property
    def f0(self):
        """Return the capacitor self resonant frequency in Hz."""
```

src/pyedb/component_libraries/ansys_components.py
```diff
@@ -33,16 +33,61 @@ def __init__(self, name, index, sbin_file):
...
+        if self.type == "Capacitor":
+            z11 = 1 / self.s_parameters.y[:, 0, 0]
+            fo_index = np.where(np.abs(z11) == np.min(np.abs(z11)))[0][0]
+            return self.s_parameters.f[fo_index]
+        else:
+            return None
+
+    @property
+    def esl(self):
+        """Returns equivalent serial inductor for capacitor only."""
```
SMoraisAnsys, line 68:
```suggestion
        """Return the equivalent serial inductor for capacitor only."""
```

src/pyedb/component_libraries/ansys_components.py
```diff
@@ -33,16 +33,61 @@ def __init__(self, name, index, sbin_file):
...
+        if self.type == "Capacitor":
+            omega_r = 2 * np.pi * self.f0
+            return 1 / (np.power(omega_r, 2) * self.cap_value)
+        else:
+            return 0.0
+
+    @property
+    def cap_value(self):
+        """Returns capacitance value."""
```
SMoraisAnsys, line 77:
```suggestion
        """Returns the capacitance value."""
```

src/pyedb/component_libraries/ansys_components.py
```diff
@@ -33,16 +33,61 @@ def __init__(self, name, index, sbin_file):
...
+        if self.type == "Capacitor":
+            return round(np.imag(self.s_parameters.y[0, 0, 0]) / (2 * np.pi * self._s_parameters.f[0]), 15)
+        else:
+            return 0.0
+
+    @property
+    def ind_value(self):
+        """Returns inductance value."""
```
SMoraisAnsys, line 85:
```suggestion
        """Return the inductance value."""
```

SMoraisAnsys (commented):
Could the processing performed be generalized and not applied only to capacitor ?
Otherwise it might be worth to have this associated to a Transistor dedicated class.

svandenb-dev:
> Could the processing performed be generalized and not applied only to capacitor ? Otherwise it might be worth to have this associated to a Transistor dedicated class.

This is only for Capacitors

PR #734, opened 2024-08-20
DOCS: Fix doc style

SMoraisAnsys:
This PR aims at fixing the documentation style issue that we are currently stacking on the repo.

On top of it, it will convert any new warning to the google rules as an error.
This way, we are sure that external contributors as well as dependabot won't have the CICD failing due to the code-style action while they didn't do any doc modification at all. This is related to the scope of each PR.


svandenb-dev (approved):
LGTM

Issue #736, opened 2024-08-21
Got negative solder balls diameters on some projects


svandenb-dev:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

For some reasons EDB api is returning sometimes negative solder ball diameter. This can be resolved by taking absolute value.

### Steps To Reproduce

NA

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

NA

PR #737, opened 2024-08-21
FIX: Set solder ball fix with negative diameter

svandenb-dev:
For some cases EDB api is returning negative pad dimension causing solder balls settings failure.

closing #736


src/pyedb/dotnet/edb_core/components.py
```diff
@@ -871,7 +871,6 @@ def create_port_on_pins(self, refdes, pins, reference_pins, impedance=50.0, port
                 elif "-" in ref_pin_name and ref_pin_name.split("-")[1] in refdes_pins:
                     ref_cmp_pins.append(refdes_pins[ref_pin_name.split("-")[1]])
             if not ref_cmp_pins:
-                self._logger.warning("No reference pins found during port creation. Port is not defined.")
```
SMoraisAnsys, line 874: Why do you remove this warning ? Can you make it a debug warning if you don't want people to see it ?

src/pyedb/dotnet/edb_core/components.py
```diff
@@ -2071,8 +2070,8 @@ def set_solder_ball(
             pin1 = list(cmp.pins.values())[0].pin
             pin_layers = pin1.GetPadstackDef().GetData().GetLayerNames()
             pad_params = self._padstack.get_pad_parameters(pin=pin1, layername=pin_layers[0], pad_type=0)
-            _sb_diam = min([self._get_edb_value(val).ToDouble() for val in pad_params[1]])
-            sball_diam = _sb_diam
+            _sb_diam = abs(min([self._get_edb_value(val).ToDouble() for val in pad_params[1]]))
+            sball_diam = 0.8 * _sb_diam
```
SMoraisAnsys, line 2074: Why do you multiply by 0.8 ?

SMoraisAnsys (commented):
Are you sure that returning the absolute value is enough ? Couldn't we have cases like [-0.6, 0.3] where 0.3 should be the value returned and this would return 0.6 ?

src/pyedb/dotnet/edb_core/components.py
svandenb-dev: this is an arbitrary value based on experiment. When users export to 3D modeler having the same value cause trouble with meshing. Using a smaller value prevent this issue

PR #746, opened 2024-08-22
FIX: Set solder ball correction

svandenb-dev:
Minor correction


src/pyedb/dotnet/edb_core/components.py
```diff
@@ -872,22 +875,21 @@ def create_port_on_pins(
...
             ref_cmp_pins = []
             for ref_pin_name in reference_pins:
                 if ref_pin_name in refdes_pins:
                     ref_cmp_pins.append(refdes_pins[ref_pin_name])
-                elif "-" in ref_pin_name and ref_pin_name.split("-")[1] in refdes_pins:
-                    ref_cmp_pins.append(refdes_pins[ref_pin_name.split("-")[1]])
+                elif "-" in ref_pin_name:
+                    if ref_pin_name.split("-")[1] in refdes_pins:
+                        ref_cmp_pins.append(refdes_pins[ref_pin_name.split("-")[1]])
             if not ref_cmp_pins:
```
ring630, line 889: Should it raise Error instead of error in log. I am afraid no one will check the log unless the code breaks.

SMoraisAnsys: Depends on what side you want to side on. Either providing something "resilient" or that breaks when it fails.
As a scientist (and project maintainer) I'd go with the second approach.
I prefer users (or myself) being notified early that something wrong is going on than deducing wrong stuff because I didn't notice a wrong behavior in the process.

svandenb-dev: Like we spoke lrt's keep it like this for now but I understand your point.

SMoraisAnsys (approved):

PR #752, opened 2024-08-27
FIX: IPC export efficiency

ring630:
Revert small changes from primitives refactoring.

primitive.layer.name is a lot slower than primitive._edb_obejct.GetLayer().GetName(). primitive.layer triggers layer collection refresh every single call.

### DIFF
--- a/src/pyedb/dotnet/edb_core/stackup.py
+++ b/src/pyedb/dotnet/edb_core/stackup.py
@@ -1800,7 +1800,7 @@ class Stackup(LayerCollection):
         temp_data = {name: 0 for name, _ in self.signal_layers.items()}
         outline_area = 0
         for i in self._pedb.modeler.primitives:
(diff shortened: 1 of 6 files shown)


src/pyedb/dotnet/edb_core/stackup.py
```diff
@@ -306,12 +304,20 @@ def layers(self):
         """
         return {name: obj for name, obj in self.all_layers.items() if obj.is_stackup_layer}
 
+    def find_layer_by_name(self, name: str):
+        """Finds a layer in a layer with given name."""
```
SMoraisAnsys, line 308:
```suggestion
        """Finds a layer with given name."""
```

SMoraisAnsys (approved):
LGTM, nice improvement of performance!

SMoraisAnsys (approved):

PR #759, opened 2024-08-30
FIX: Arbitrary wave port

svandenb-dev:


SMoraisAnsys (changes requested):
To avoid adding too many lines of code, could you please define a variable to stock `"add_on_top"` or `"add_on_bottom"` and only call twice the `add_layer` method ?
This would avoid quasi duplicate content on `if` and `else` statements.

src/pyedb/dotnet/edb.py
```diff
@@ -4462,25 +4468,29 @@ def create_model_for_arbitrary_wave_ports(
...
+            base_layer="ports",
+        )
+        cloned_edb.stackup.add_layer(
+            layer_name="port_pec",
+            layer_type="signal",
+            thickness=launching_box_thickness,
+            method="add_on_top",
+            material="pec",
+            base_layer="ports",
+        )
```
SMoraisAnsys, line 4493:
```suggestion
        cloned_edb.stackup.add_layer(
            layer_name="ref",
            layer_type="signal",
            thickness=0.0,
            material="pec",
            method=f"add_on_{mounting_side}",
            base_layer="ports",
        )
        cloned_edb.stackup.add_layer(
            layer_name="port_pec",
            layer_type="signal",
            thickness=launching_box_thickness,
            method=f"add_on_{mounting_side}",
            material="pec",
            base_layer="ports",
        )
```

SMoraisAnsys: New changes discarded the option to add on bottom. Since you added a check on the value of `mounting_side`, we can leverage it in an f-string.

SMoraisAnsys (changes requested):

src/pyedb/dotnet/edb.py
```diff
@@ -4462,25 +4468,29 @@ def create_model_for_arbitrary_wave_ports(
...
-        if mounting_side == "top":
-            cloned_edb.stackup.add_layer(
-                layer_name="port_pec", layer_type="signal", thickness=box_thick, method="add_on_bottom", material="pec"
-            )
-        else:
-            cloned_edb.stackup.add_layer(
-                layer_name="port_pec", layer_type="signal", thickness=box_thick, method="add_on_top", material="pec"
-            )
-
+        cloned_edb.stackup.add_layer(
```
SMoraisAnsys, line 4478:
```suggestion
        if launching_box_thickness:
            launching_box_thickness = self.edb_value(launching_box_thickness).ToString()
        cloned_edb.stackup.add_layer(
```

SMoraisAnsys (changes requested):
Adding last comment to ensure that user's `launching_box_thickness` is correctly handled.

SMoraisAnsys (approved):

PR #763, opened 2024-09-02
FIX: version property, fix bug in loac configuration, fix bug in quit_application

ring630:


src/pyedb/siwave.py
```diff
@@ -22,6 +23,7 @@
 from pyedb.generic.general_methods import _pythonver, is_windows
 from pyedb.misc.misc import list_installed_ansysem
 from pyedb.siwave_core.icepak import Icepak
+from tests.conftest import generate_random_string
```
SMoraisAnsys, line 26: Please avoid using tests methods in the code of the project.

SMoraisAnsys (changes requested):

src/pyedb/siwave.py
```diff
@@ -11,11 +11,14 @@
 import os
 from pathlib import Path
 import pkgutil
+import shutil
 import sys
 import tempfile
 import time
 import warnings
 
+from ansys.aedt.core import generate_unique_name
```
SMoraisAnsys, line 20:
```suggestion
from pyedb.generic.general_methods import generate_unique_name
```

SMoraisAnsys: You cannot use `pyaedt` here as it is not a default dependency.

src/pyedb/siwave.py
```diff
@@ -509,16 +518,17 @@ def load_configuration(self, file_path: str):
...
         self.export_edb(temp_edb)
         self.oproject.ScrCloseProject()
-        edbapp = Edb(temp_edb, edbversion=self.current_version)
+        edbapp = Edb(temp_edb, edbversion=self.version)
         edbapp.configuration.load(file_path)
         edbapp.configuration.run()
         edbapp.save()
         edbapp.close()
         self.import_edb(temp_edb)
+        shutil.rmtree(Path(temp_edb))
```
SMoraisAnsys, line 531:
```suggestion
        shutil.rmtree(Path(temp_edb), ignore_errors=True)
```
This will avoid test failure ok cleanup (if that ever happens).

SMoraisAnsys (changes requested):

SMoraisAnsys (approved):

PR #773, opened 2024-09-05
REFACTOR: improve edb.nets.plot() efficiency

ring630:
There is a significant efficiency impact on edb.nets.plot() from version 0.6.0. This refactoring fixed this efficiency issue.
image

### DIFF
--- a/src/pyedb/dotnet/edb_core/nets.py
+++ b/src/pyedb/dotnet/edb_core/nets.py
@@ -485,8 +485,8 @@ class EdbNets(object):
         for path in self._pedb.modeler.paths:
             if path.is_void:
                 continue
-            net_name = path.net_name
-            layer_name = path.layer_name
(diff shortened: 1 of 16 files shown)


src/pyedb/dotnet/edb_core/nets.py
```diff
@@ -531,8 +531,8 @@ def get_plot_data(
         for poly in self._pedb.modeler.polygons:
             if poly.is_void:
                 continue
-            net_name = poly.net_name
-            layer_name = poly.layer_name
+            net_name = path.net_name
```
svandenb-dev, line 534: Why changing poly to path ?

ring630: It is a mistake.

svandenb-dev (approved):
LGTM

PR #781, opened 2024-09-09
FIX: Replace pyaedt to ansys.aedt.core

Samuelopez-ansys:
The one in settings was used in PyAEDT when PyEDB is initialized, so the warning was appearing in PyAEDT. 

I also changed the examples to avoid the warnings.


svandenb-dev (approved):
LGTM

PR #782, opened 2024-09-11
FEAT: Grpc transition

svandenb-dev:


hui-zhou-a:
> @svandenb-dev I won't have the time to review the changes with my current work load. @hui-zhou-a Can you have a look at that ?

@svandenb-dev is this PR ready for review? Some tests are failing.

PR #792, opened 2024-09-16
FIX: padstack and siwave

### LINKED ISSUE #793: Bug located in padstack instance

### Description of the bug

Cache pad_by_layer doesn't get updated when layer name is changed. 

### Steps To Reproduce

Change layer name, and then call pad_by_layer

Windows

3.7

### Installed packages

all

ring630:
- bug fix #793 
- Better file path management in Siwave class
- typo fix in example
- Minor fix in configuration class

### DIFF
--- a/tests/legacy/system/test_edb_configuration_2p0.py
+++ b/tests/legacy/system/test_edb_configuration_2p0.py
@@ -386,9 +386,7 @@ class TestClass:
 
     def test_06_s_parameters(self, edb_examples):
         data = {
-            "general": {
-                "s_parameter_library": self.local_input_folder
-            },
+            "general": {"s_parameter_library": self.local_input_folder},
             "s_parameters": [
                 {
                     "name": "GRM32_DC0V_25degC_series",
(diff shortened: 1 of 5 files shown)


tests/legacy/system/test_edb_configuration_2p0.py
```diff
@@ -385,10 +385,28 @@ def test_05f_ports_between_two_points(self, edb_examples):
...
+                    "name": "GRM32_DC0V_25degC_series",
+                    "file_path": "GRM32_DC0V_25degC_series.s2p",
+                    "apply_to_all": False,
+                    "component_definition": "CAPC3216X190X55ML30T25",
+                    "components": ["C59"],
+                    "reference_net": "GND",
+                    "reference_net_per_component": {"C59": "GND"},
+                },
+            ],
+        }
```
SMoraisAnsys, line 409: Why are you defining data manually instead of using `s_parameter.json` like before ?

SMoraisAnsys (approved):
LGTM, I left a minor comment.

tests/legacy/system/test_edb_configuration_2p0.py
ring630: This is for the sake of debugging. If something goes wrong, I can check everything in one place instead of jumping between different files.

Issue #800, opened 2024-09-19
Version 0.28 break pyaedt CICD


SMoraisAnsys:
The latest release (v0.28.0) was breaking `pyaedt` CICD when running solver test in Linux.

To avoid being totally blocked, `pyaedt` is currently blocking this version (https://github.com/ansys/pyaedt/pull/5181).

SMoraisAnsys:
@svandenb-dev @ring630 Any idea on what could cause this sudden problem ?

SMoraisAnsys:
Recent changes seem to have fixed the problem, see https://github.com/ansys/pyaedt/pull/5184.
In particular, it seems that this PR fixed the issue https://github.com/ansys/pyedb/pull/808.
However, I'll let this issue open as there is a need to dig into what was exactly the root of the problem.

SMoraisAnsys:
THe issue seem to be related to `dotnetcore2`, we should avoid using this old package and move to another approach where users have to install `dotnet` / `mono` by themselves.

PR #801, opened 2024-09-19
CI: Add pyaedt linux testing when releasing

SMoraisAnsys:
To avoid as much as possible a situation where pyaedt's CICD is no longer working after a release of pyedb, see #800, I'd like to add a job to tests pyaedt in Linux.


ring630 (approved):
LGTM

PR #804, opened 2024-09-19
FIX: Fix disjoint_nets method

maxcapodi78:
area method is failing for PadstackInstances
the net_list argument is already used in the same loop. Renamed to avoid issues


ring630 (approved):
LGTM

svandenb-dev (approved):
LGTM

PR #810, opened 2024-09-23
FEAT: EDB Config - export component properties

ring630:


SMoraisAnsys (commented):
Could the method added to components by made into python properties instead of get/set calls ? Btw, is there a reason to remove the classes `CfgXXXProperties` ?

ring630:
> Could the method added to components by made into python properties instead of get/set calls ? Btw, is there a reason to remove the classes `CfgXXXProperties` ?
get/set is replaced by property in the coming commit. 

I do not find the value of CfgXXXProperties. The less the better. I will bring them back when we need them.

svandenb-dev (approved):
LGTM

svandenb-dev (approved):
LGTM

PR #817, opened 2024-09-25
FIX: Speedup Edb plot method

maxcapodi78:
Speed up the edb plot method by reusing the layer color properties.


svandenb-dev (approved):
LGTM

PR #824, opened 2024-09-27
REFACTOR: edb configuration stackup layer

hui-zhou-a:


src/pyedb/configuration/cfg_setup.py
```diff
@@ -177,6 +177,9 @@ def apply(self):
     def get_data_from_db(self):
         setups = []
         for _, s in self._pedb.setups.items():
+            if not s.type == "hfss":
+                # todo exporting siwave setup is skipped due to defect in 24.2
```
SMoraisAnsys, line 181: Can you log a warning here to notify of such behavior ? Also, it might be a good idea to extend the if statement with a check on AEDT version, right ?

src/pyedb/dotnet/edb_core/edb_data/layer_data.py
```diff
@@ -238,6 +238,26 @@ def type(self, value):
...
+    def properties(self, params):
+        name = params.get("name")
+        if name:
+            self.name = name
+        type = params.get("type")
+        if type:
+            self.type = type
+        color = params.get("color")
+        if color:
+            self.color = color
```
SMoraisAnsys, line 259:
```suggestion
        name = params.get("name", "")
        if name:
            self.name = name
        type = params.get("type", "")
        if type:
            self.type = type
        color = params.get("color", "")
        if color:
            self.color = color
```
Adding default value will ensure that if someone does not provide one of the expected parameters, this call will not raise an exception. However, if you want an exception to be triggered because you want users to provide parameters, discard my changes.

src/pyedb/dotnet/edb_core/edb_data/layer_data.py
```diff
@@ -705,3 +725,35 @@ def _load_layer(self, layer):
...
+        data["name"] = self.name
+        data["type"] = self.type
+        data["material"] = self.material
+        data["fill_material"] = self.fill_material
+        data["thickness"] = self._edb_object.GetThicknessValue().ToString()
+        data["color"] = self.color
+        return data
+
+    @properties.setter
+    def properties(self, params):
```
SMoraisAnsys, line 741: Same remark as before.

tests/legacy/system/test_edb_configuration_2p0.py
```diff
@@ -647,7 +647,7 @@ def test_12_setup_siwave_dc(self, edb_examples):
         }
         edbapp = edb_examples.get_si_verse()
         assert edbapp.configuration.load(data, apply_file=True)
-        data_from_db = edbapp.configuration.get_data_from_db(setups=True)
+        """data_from_db = edbapp.configuration.get_data_from_db(setups=True)
```
SMoraisAnsys, line 650: Something is wrong here. Either remove the code or "uncomment" it.

tests/legacy/system/test_edb_configuration_2p0.py
```diff
@@ -839,14 +837,14 @@ def test_14_setup_siwave_syz(self, edb_examples):
...
```
SMoraisAnsys, line 840: Same remark as previous comment.

SMoraisAnsys (changes requested):

src/pyedb/configuration/cfg_setup.py
hui-zhou-a: Done

src/pyedb/dotnet/edb_core/edb_data/layer_data.py
hui-zhou-a: dict.get() returns None if key doesn't exist, so no exception will be triggered.

src/pyedb/dotnet/edb_core/edb_data/padstacks_data.py
```diff
@@ -499,7 +499,7 @@ def pad_parameters(self) -> dict:
             self._ppadstack._pedb._edb.Definition.PadType.RegularPad,
             self._ppadstack._pedb._edb.Definition.PadType.AntiPad,
             self._ppadstack._pedb._edb.Definition.PadType.ThermalPad,
-            self._ppadstack._pedb._edb.Definition.PadType.Hole,
+            # self._ppadstack._pedb._edb.Definition.PadType.Hole,
```
SMoraisAnsys, line 502: Is that a miss or on purpose ? If on purpose can you comment why and resolve this comment ?

tests/legacy/system/test_edb_configuration_2p0.py
SMoraisAnsys: This is marked as resolved but the code is still here. Should it be removed ?

src/pyedb/dotnet/edb_core/edb_data/padstacks_data.py
svandenb-dev: @hui-zhou-a can you please check this comment so we can merge the PR ?

hui-zhou-a: Added comments.
# This property doesn't appear in UI. It is unclear what it is used for. Suppressing this property for now.

tests/legacy/system/test_edb_configuration_2p0.py
hui-zhou-a: I thought I removed them. Now they are removed.

SMoraisAnsys (approved):
LGTM

PR #826, opened 2024-09-27
FIX: import_gds_file

gfetsis:


svandenb-dev (dismissed):
LGTM

src/pyedb/dotnet/edb.py
```diff
@@ -1484,31 +1480,67 @@ def import_gds_file(
...
+            temp_map_file = os.path.splitext(inputGDS)[0] + ".map"
+            temp_layermap_file = os.path.splitext(inputGDS)[0] + ".layermap"
+
+            if map_file is None:
+                if os.path.isfile(temp_map_file):
+                    map_file = temp_map_file
+                elif os.path.isfile(temp_layermap_file):
+                    map_file = temp_layermap_file
+                else:
+                    print("No layermap file is provided!")
```
SMoraisAnsys, line 1518:
```suggestion
                    self.logger.error("Unable to define map file.")
```

src/pyedb/dotnet/edb.py
```diff
@@ -1484,31 +1480,67 @@ def import_gds_file(
...
+
+            if tech_file is None:
+                if control_file is None:
+                    temp_control_file = os.path.splitext(inputGDS)[0] + ".xml"
+                    if os.path.isfile(temp_control_file):
+                        control_file = temp_control_file
+                    else:
+                        print("No control or tech file is provided!")
```
SMoraisAnsys, line 1526:
```suggestion
                        self.logger.error("Unable to define control file.")
```

SMoraisAnsys (commented):
It would be better to log instead of printing.
Also, if it doesn't make sense to run the command later on, I'd raise an exception `RuntimeError`.

gfetsis:
@SMoraisAnsys I committed the proposed changes. As long as there is an input GDS (even without map or tech file), the translator should be ran so I didn't add the exception RuntimeError. Let me know if that makes sense.

svandenb-dev (approved):
LGTM great job

PR #840, opened 2024-10-04
BUILD: Remove dotnetcore2 dependency

SMoraisAnsys:
To be added in a near future.
Follow up of #839, the documentation should be updated.


svandenb-dev (approved):
LGTM

PR #845, opened 2024-10-09
FIX: Improve cutout speed

maxcapodi78:


svandenb-dev (approved):
LGTM

src/pyedb/dotnet/edb_core/cell/primitive/primitive.py
```diff
@@ -62,6 +60,14 @@ def __init__(self, pedb, edb_object):
...
         }
 
+    @property
+    def _core_stackup(self):
+        return self._app.stackup
+
+    @property
+    def _core_net(self):
+        return self._app.nets
+
```
hui-zhou-a, line 70: I suggest to remove these two properties

self._core_stackup is only used in layer_name()

self._core_net is never used.

hui-zhou-a, line 70:
```suggestion

```

hui-zhou-a: replace layer_name setter with 

```
layers = self._pedb.stackup.layers
        if isinstance(val, str) and val in layers.keys():
            layer = layers[val]._edb_layer
            if layer:
                self.primitive_object.SetLayer(layer)
            else:
                raise AttributeError("Layer {} not found.".format(val))
        elif isinstance(val, type(layers[list(layers.keys())[0]])):
            try:
                self.primitive_object.SetLayer(val._edb_layer)
            except:
                raise AttributeError("Failed to assign new layer on primitive.")
        else:
            raise AttributeError("Invalid input value")
```

maxcapodi78: I think solution 1 is more elegant. It doesn't impact on speed and there is only 1 place where the layer name is setup. Furthermore solution 2 requires much more redundant line codes

maxcapodi78: core.nets is not used in Primitive but that class is inherited in many others and there should be a need to use it

hui-zhou-a: Oh, yes. Let's keep it.

hui-zhou-a (approved):
LGTM.

PR #849, opened 2024-10-09
FEAT: hole override

hui-zhou-a:


svandenb-dev (approved):
LGTM

SMoraisAnsys (approved):
LGTM
As you mentioned in a comment of a previous PR, `get` returns `None` by default. We might want to update the code to avoid `.get("...", None)` pattern.

PR #854, opened 2024-10-16
FIX: setup types

hui-zhou-a:
PyEDB is failing on 24R1

### DIFF
--- a/src/pyedb/dotnet/edb.py
+++ b/src/pyedb/dotnet/edb.py
@@ -3612,15 +3612,15 @@ class Edb(Database):
         setup_types = {}
         setups = {}
         for i in list(self.active_cell.SimulationSetups):
-            if i.GetType().endswith(".kHFSS"):
+            if i.GetType().ToString().endswith(".kHFSS"):
                 setups[i.GetName()] = HfssSimulationSetup(self, i)
-            elif i.GetType().endswith(".kSIWave"):
+            elif i.GetType().ToString().endswith(".kSIWave"):
                 setups[i.GetName()] = SiwaveSimulationSetup(self, i)
-            elif i.GetType().endswith(".kSIWaveDCIR"):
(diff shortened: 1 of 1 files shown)


src/pyedb/dotnet/edb.py
```diff
@@ -3609,17 +3609,18 @@ def setups(self):
         Dict[str, :class:`legacy.edb_core.edb_data.siwave_simulation_setup_data.SiwaveSYZSimulationSetup`]
 
         """
+        setup_types = {}
```
SMoraisAnsys, line 3612:
```suggestion
```

SMoraisAnsys (commented):
I'm not sure to understand what was failing in 2024.1.
Is it because some `SimulationSetupType.whateverthename` do not exist in that version ?
I left a minor comment due to an unused variable.

hui-zhou-a:
> I'm not sure to understand what was failing in 2024.1. Is it because some `SimulationSetupType.whateverthename` do not exist in that version ? I left a minor comment due to an unused variable.

That's right. below property doesn't exist in 2024.1.

`self.edb_api.utility.utility.SimulationSetupType.kRaptorX`

SMoraisAnsys:
> > I'm not sure to understand what was failing in 2024.1. Is it because some `SimulationSetupType.whateverthename` do not exist in that version ? I left a minor comment due to an unused variable.
> 
> That's right. below property doesn't exist in 2024.1.
> 
> `self.edb_api.utility.utility.SimulationSetupType.kRaptorX`

If there is no changes associated to the "name" of the SimulationSetupType in coming AEDT releases then this changes will be great. Otherwise, we might want to keep the previous approach and add extra if condition based on AEDT versions. For example, we could have an if statement like `<=2024.1` before the test on kRaptorX.

SMoraisAnsys (approved):

PR #856, opened 2024-10-17
CHORE: Temporary removal of docformatter hook

SMoraisAnsys:
As title says.

The reason for this change is related to a new major release of precommit where `python_venv` configuration can no longer be used. Since the docformatter is still using it ATM, we should temporarily comment this check.
Note that https://github.com/PyCQA/docformatter/pull/287 fixes the problem and has been merged last week. We are just waiting for a new release.


svandenb-dev (dismissed):
LGTM

svandenb-dev (approved):
LGTM

PR #862, opened 2024-10-19
FIX: via name convention

hui-zhou-a:
make the via naming convention the same as AEDT


SMoraisAnsys (approved):
Can you tell us where this approach is used in PyAEDT ?

PR #863, opened 2024-10-19
REFACTOR: EDB configuration

hui-zhou-a:


hui-zhou-a:
@svandenb-dev I move properties methods from padstack and component to Configuration class. I am keeping layer.properties as it is since layer class is completely different in grpc. You don't need to refactor this method.

@SMoraisAnsys, the properties methods are gone now. I modified the unittest you mentioned on Friday.

hui-zhou-a:
@svandenb-dev properties methods from padstack and component are moved to configuration class. I am keeping properties method in stackup since stackup class is completely different in grpc. There is no need to refactor it.

@SMoraisAnsys properties methods are gone. I modified the unittest you wrote.

svandenb-dev (approved):
LGTM

PR #868, opened 2024-10-21
FEAT: Add default value to material permeability

### LINKED ISSUE #827: The default permeability value of new material cannot be 0

### Description of the bug

The default permeability of new material is should 1.

### Steps To Reproduce

add_material

Windows

3.7

### Installed packages

all

SMoraisAnsys:
Ensure that users creating material through `pyedb` end up with a material with a permeability equal to `1` instead of `0` (default value retrieved through EDB API).

Closes #827


hui-zhou-a (approved):
LGTM

PR #880, opened 2024-11-01
FIX: polygon doesn't update after expansion

hui-zhou-a:
- expand method from .net api create a new polygon. It needs to be set to the primitive.
- only polygon can expand. Move expand method to EdbPolygon class
- Remove uittest for Path
- do not export padstack definition by default.


svandenb-dev (approved):
LGTM

PR #881, opened 2024-11-04
FEAT: add export gds comps xml function

felipeescastro:


src/pyedb/dotnet/edb.py
```diff
@@ -4598,3 +4598,51 @@ def definitions(self):
...
+
+        if not comps_to_export:
+            comps_to_export = self.components.components
+
+        for comp in comps_to_export:
+            ocomp = self.components.components[comp]
+            fo_txt += "\t<GDS_COMPONENT>\n"
+            for pin in ocomp.pins:
+                pins_position_unit = unit_converter(ocomp.pins[pin].position, output_units=gds_comps_unit)
+                fo_txt += '\t\t<GDS_PIN Name="{}" x="{}" y="{}" Layer="{}"/>\n'.format(
```
svandenb-dev, line 4631: I would use xml.etree instead of working on raw string

svandenb-dev (approved):
LGTM

SMoraisAnsys (approved):

felipeescastro:
> @felipeescastro Could you see if the test isn't taking too long to run ? It seems that the time required to run the CICD (for tests) has doubled in this PR.

@SMoraisAnsys I'm not sure. The code runs smoothly locally and the unittest I added is supposed to be as fast as any other test in the test file.

SMoraisAnsys:
> > @felipeescastro Could you see if the test isn't taking too long to run ? It seems that the time required to run the CICD (for tests) has doubled in this PR.
> 
> @SMoraisAnsys I'm not sure. The code runs smoothly locally and the unittest I added is supposed to be as fast as any other test in the test file.

Simon fixes the bug, have a look at his changes :)

Issue #889, opened 2024-11-12
Bug located in edb cfg example


hui-zhou-a:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

Below two methods don't show plots in documentation. 
1, edbapp.nets.plot
![Image](https://github.com/user-attachments/assets/5f94a497-181f-4f6f-9628-a7fe539f53a3)

2,h3d.post.get_solution_data(expressions="mag(S(DIFF_CONN,DIFF_BGA))", context="Differential Pairs").plot
![Image](https://github.com/user-attachments/assets/5aab4404-8d24-489b-b8c1-5b16c3b75e81)

### Steps To Reproduce

https://edb.docs.pyansys.com/version/dev/examples/use_configuration/serdes.html

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

all

svandenb-dev:
@hui-zhou-a can we close this issue ?

PR #892, opened 2024-11-12
DOCS: Correct typos in class names/paths

SMoraisAnsys:
@hui-zhou-a While working on a fix to #889 I found typos in class references. This should patch those.


hui-zhou-a (approved):
LGTM

PR #899, opened 2024-11-16
FIX: read Spice model from file

hui-zhou-a:


SMoraisAnsys (commented):
TBH I don't think we should be handling the encoding issue in `pyedb`. The problem being solved is that a file was written with an encoding that is incompatible with Linux. Is it something that we are creating and then can't read ourselves ? If that's the case, then we need to deal with it to ensure that a generated file can be read in the same space. However, if the file was written using Windows encoding, I would prefer users to perform the conversion to make it compatible with Linux (`iconv -f "my-encoding" -t utf-8 origine_file reencoded_file`).

Otherwise, we'll have to add a dependency that is not very used (could add security issues) and we would have to add this changes in every part of the code base where a file is open.

aohta2021:
I  can get successfule on Linux after changed to UTF-8 format of tcm0403s_350_2p_t210_s.mod( SPICE file).
The issue was SPICE format in Linux environment.
It have to changed to UTF-8 format.
This issue has been fixed.

Thank you.

PR #901, opened 2024-11-19
REFACTOR: Material to match gRPC implementation

SMoraisAnsys:
Following a discussion on how gRPC's `Material` is currently implemented.

> Note: prior to this changes, the implementation was leveraging a `__property` attribute which was updated every time a property, like `conductivity`, was accessed. However, the gRPC doesn't have such attribute so it does not make sense to keep it in dotnet.


svandenb-dev (approved):
LGTM

svandenb-dev (approved):
LGTM

PR #903, opened 2024-11-20
REFACTOR: New edb cfg example dcir

hui-zhou-a:


examples/use_configuration/pcb_dc_ir.py
```diff
@@ -0,0 +1,174 @@
...
+    show_region=True,
+    show_axis=True,
+    show_grid=True,
+    show_ruler=True,
+)
+
+# ## Compute power loss
+
+p_layers = h3d.post.compute_power_by_layer(layers=["1_Top"])
+display(p_layers)
```
SMoraisAnsys, line 164: @hui-zhou-a  Is there a reason to use `IPython.display` here ? Don't we have internal method to display the layers ?
Note that the associated image is note displayed in the HTML pages.

examples/use_configuration/pcb_dc_ir.py
```diff
@@ -0,0 +1,174 @@
...
+from IPython.display import display
+from ansys.aedt.core import Hfss3dLayout
+from ansys.aedt.core.downloads import download_file
+
+from pyedb import Edb
+
+AEDT_VERSION = "2024.2"
+NG_MODE = False
+
+# -
```
SMoraisAnsys, line 22:
```suggestion
# +
import json
import os
import tempfile

from IPython.display import display
from ansys.aedt.core import Hfss3dLayout
from ansys.aedt.core.downloads import download_file

from pyedb import Edb

AEDT_VERSION = "2024.2"
NG_MODE = False

# -
```

SMoraisAnsys (approved):
LGTM, I left minor comments

examples/use_configuration/pcb_dc_ir.py
hui-zhou-a: @SMoraisAnsys could you give me an example of the internal method?

hui-zhou-a: @SMoraisAnsys how to show a figure then?

hui-zhou-a: I think it is an issue in PyAEDT. It doesn't plot automatically. I will investigate it in another branch.

SMoraisAnsys (dismissed):
LGTM, I left a minor comment

svandenb-dev (approved):
LGTM

PR #916, opened 2024-11-27
REFACTOR: Dotnet coreclr loading

SMoraisAnsys:
The changes allow one to leverage a dotnet installation at system level.
This is something that we should encourage as using `dotnetcore2` prevents us from providing a conda recipe and also because `dotnetcore2` requires openssl1.1 to be installed. It can't work with newer versions.


SMoraisAnsys:
The pyaedt tests failed on linux when running with pyaedt-ci-6. I will conduct some changes on the VM to ensure it behaves correctly following our warnings.

svandenb-dev (approved):
LGTM

PR #924, opened 2024-12-02
FIX: DOTNET_ROOT handling with pyaedt

### LINKED ISSUE #922: Bug located in clr_module

### Description of the bug

Following #916 changes, it seems that one might face a problem when using dotnetcore2 and pyaedt.
Indeed, pyaedt has it's own clr_module and loads dotnetcore2 before pyedb does. Therefore, DOTNET_ROOT is set to the python environment dotnetcore2 location. This is a problem since the current implementation leads to loading a configuration file that isn't present.

### Steps To Reproduce

Image

Windows

3.10

### Installed packages

Not relevant

SMoraisAnsys:
As title says, see #922 for more information.

Below is the result after calling the same sequences that caused an error previously:

image

Close #922

### DIFF
--- a/src/pyedb/dotnet/clr_module.py
+++ b/src/pyedb/dotnet/clr_module.py
@@ -48,6 +48,7 @@ if is_linux:  # pragma: no cover
             runtime = get_coreclr()
             load(runtime)
             print(runtime)
+            print(runtime.dotnet_root)
             # os.environ["DOTNET_ROOT"] = runtime.dotnet_root
             is_clr = True
         # TODO: Fall backing to dotnetcore2 should be removed in a near future.
--- a/src/pyedb/dotnet/clr_module.py
+++ b/src/pyedb/dotnet/clr_module.py
@@ -4,8 +4,6 @@ import pkgutil
 import sys
 import warnings
 
-from clr_loader import find_runtimes
-
 import pyedb
 
 LINUX_WARNING = (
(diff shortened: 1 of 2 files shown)


svandenb-dev (dismissed):
LGTM

svandenb-dev (approved):
LGTM

src/pyedb/dotnet/clr_module.py
```diff
@@ -34,79 +33,62 @@ def custom_show_warning(message, category, filename, lineno, file=None, line=Non
...
                     "Configuration file could not be found from DOTNET_ROOT. "
                     "Please ensure that .NET SDK is correctly installed or "
                     "that DOTNET_ROOT is correctly set."
                 )
+            runtime_config = candidates[0]
+    # Use specific .NET core runtime
+    if dotnet_root is not None and runtime_config is not None:
         try:
             load("coreclr", runtime_config=str(runtime_config), dotnet_root=str(dotnet_root))
+            os.environ["DOTNET_ROOT"] = dotnet_root
```
isaacwaldron, line 86: This line will fail if we have entered the except block at line 51 to fall back to dotnetcore2 as dotnet_root is set to a Path object on line 60 and we do not convert to str here (though we should probably use `Path.as_posix()` instead).  I believe this is occurring on the toolkit test runners if we do not initialize PyAEDT first (thus ensuring that DOTNET_ROOT is already set).  I can provide a link to a run demonstrating this.

PR #933, opened 2024-12-05
FIX: Set DOTNET_ROOT with str and not Path

SMoraisAnsys:
As pointed out in https://github.com/ansys/pyedb/pull/924/files#r1870405256, the value that we are passing to `os.environ` is a `Path` instance while it should be `str`. This solves this issue.

@isaacwaldron I'll check that this works correctly when I get back from my trip and make a new release.

### DIFF
--- a/tests/legacy/unit/test_clr_module.py
+++ b/tests/legacy/unit/test_clr_module.py
@@ -1,0 +1,92 @@
+import os
+from pathlib import Path
+import sys
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+DOTNET_ROOT = "dummy/root/path"
+DOTNET_ROOT_PATH = Path(DOTNET_ROOT)
+DOTNETCORE2_FILE = "dummy/dotnetcore2/file"
+DOTNETCORE2_BIN = "dummy/dotnetcore2/bin"
+PYEDB_FILE = "dummy/pyedb/file"
+
+
+@pytest.fixture
+def clean_environment():
+    if "pyedb.dotnet.clr_module" in sys.modules:
+        del sys.modules["pyedb.dotnet.clr_module"]
+    if "DOTNET_ROOT" in os.environ:
+        del os.environ["DOTNET_ROOT"]
+
+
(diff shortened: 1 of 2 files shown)


isaacwaldron (approved):
Looks good!

isaacwaldron:
@SMoraisAnsys looks good to me.  Interesting that line 86 is covered, since you haven't modified any test to do this I'm surprised this could have been passing earlier as apparently there is a workflow that activates this code path.

SMoraisAnsys:
> @SMoraisAnsys looks good to me. Interesting that line 86 is covered, since you haven't modified any test to do this I'm surprised this could have been passing earlier as apparently there is a workflow that activates this code path.

It isn't covered by testing. I'll extend this PR with some tests on clr_module

tests/legacy/unit/test_clr_module.py
```diff
@@ -0,0 +1,92 @@
...
+
+DOTNET_ROOT = "dummy/root/path"
+DOTNET_ROOT_PATH = Path(DOTNET_ROOT)
+DOTNETCORE2_FILE = "dummy/dotnetcore2/file"
+DOTNETCORE2_BIN = "dummy/dotnetcore2/bin"
+PYEDB_FILE = "dummy/pyedb/file"
+
+
+@pytest.fixture
+def clean_environment():
```
isaacwaldron, line 16: Should this fixture yield at the end and then restore the original state after?  Otherwise, these tests are potentially changing the global state for other tests.

SMoraisAnsys: In our current CICD way of working, this shouldn't be a problem but you are correct.
I'll update that right away

SMoraisAnsys:
Thanks for the review @isaacwaldron let me know if something else is not correct

isaacwaldron:
@SMoraisAnsys I had started a test run using a dependency to the initial commit (which has the actual change I need) on my downstream project that was failing.  So far so good... will come back and finally approve when that passes.

isaacwaldron (approved):
Looks good and passing in a downstream test suite that was previously failing.  Thanks for the quick turnaround!

PR #957, opened 2024-12-26
FIX: Do not raise str while only classes or instances are allowed.

MaxJPRey:


src/pyedb/configuration/cfg_ports_sources.py
```diff
@@ -302,7 +302,7 @@ def _create_terminals(self):
                         terminal_name: self._pedb.components.instances[self.reference_designator].pins[neg_value]
                     }
                 else:
-                    raise f"Wrong negative terminal type {neg_type}"
+                    raise Exception(f"Wrong negative terminal type {neg_type}")
```
MaxJPRey, line 305:
```suggestion
                    raise Exception(f"Wrong negative terminal type {neg_type}.")
```

hui-zhou-a (approved):
LGTM

PR #959, opened 2024-12-26
FIX: Property is_enabled was defined several times in the same class.

MaxJPRey:


src/pyedb/dotnet/edb_core/cell/hierarchy/component.py
```diff
@@ -346,28 +346,6 @@ def is_null(self):
...
-            return True
-
-    @is_enabled.setter
-    def is_enabled(self, enabled):
-        """Enables the current object."""
-        if self.type in ["Resistor", "Capacitor", "Inductor"]:
-            component_property = self.component_property
-            component_property.SetEnabled(enabled)
-            self.edbcomponent.SetComponentProperty(component_property)
-
```
MaxJPRey, line 370: @hui-zhou-a @svandenb-dev . I think we can delete those lines as `is_enabled` is deprecated in favor of `enabled` based on this comment:

https://github.com/ansys/pyedb/blob/3172b5b4ed64e1f3f21af9c07bf45e2c59290565/src/pyedb/dotnet/edb_core/cell/hierarchy/component.py#L183-L197

Your expertise will decide if we keep it or not 😃 .

hui-zhou-a: @svandenb-dev is it time to delete all the deprecated methods?

PR #964, opened 2025-01-03
FIX: fix padstack def

hui-zhou-a:


src/pyedb/configuration/cfg_padstacks.py
```diff
@@ -124,6 +124,15 @@ def set_parameters_to_edb(self):
         if self.solder_ball_parameters:
             self._set_solder_parameters_to_edb(self.solder_ball_parameters)
 
+    def retrieve_parameters_from_edb(self):
```
SMoraisAnsys, line 127: There is no change here and the method was just moved to avoid mixing public and protected methods, right ?

SMoraisAnsys (approved):
LGTM

src/pyedb/configuration/cfg_padstacks.py
hui-zhou-a: @SMoraisAnsys that's correct

PR #965, opened 2025-01-08
FIX: Point terminal layer getter function

SMoraisAnsys:
Duplicates PR #945 which was created from a fork.

The layer getter property of the point terminal always returned the top layer name irrespective of where the terminal was located. This has been fixed.

(cherry picked from commit 265b4b3db7487e050270a9e9a81a88538cd1501f)

### DIFF
--- a/src/pyedb/dotnet/edb_core/cell/terminal/terminal.py
+++ b/src/pyedb/dotnet/edb_core/cell/terminal/terminal.py
@@ -105,12 +105,11 @@ class Terminal(Connectable):
     @property
     def layer(self):
         """Get layer of the terminal."""
-        point_data = self._pedb.point_data(0, 0)
-        layer = list(self._pedb.stackup.layers.values())[0]._edb_layer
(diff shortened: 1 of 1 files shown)


svandenb-dev (approved):
LGTM

hui-zhou-a:
@svandenb-dev the layer property should be in PointTerminal class not Terminal

skandak-ansys:
@hui-zhou-a `layer` property is available for both `PointTerminal `and `PadstackInstanceTerminal`. So I added a `try-except` to capture terminals where layer property is invalid.

hui-zhou-a:
@skandak-ansys logger.error should be replaced by Raise RuntimeError. Logging the error will hide the problem.

hui-zhou-a:
@skandak-ansys GetParameter should only be called in PointTerminal and PadstackInstanceTermal class. This is a code structural problem. I will fix it in another PR. Thanks for your contribution!

PR #966, opened 2025-01-08
REFACTOR: Optimize Edb initialization for die designs

SMoraisAnsys:
Duplicates PR #962 which was created from a fork.

---

Several times, die designs don't have any component in them.
The code changes made in this PR signifiantly reduces the Component initialization time in this case which thereby reduces the Edb initialization time by ~50% especially when there are >100k Groups in the layout.

This should also speed up EDB initialization time when there are 1000s of components in the design (Ex: Deep trench capacitors in the IC).


hui-zhou-a (approved):
LGTM

PR #967, opened 2025-01-09
FIX: bug in create pin group, replace logger.error with raise Value e…

hui-zhou-a:
…rror

### DIFF
--- a/src/pyedb/configuration/cfg_pin_groups.py
+++ b/src/pyedb/configuration/cfg_pin_groups.py
@@ -74,7 +74,7 @@ class CfgPinGroup(CfgBase):
             comp = self._pedb.components.instances[self.reference_designator]
             pins = [p for p, obj in comp.pins.items() if obj.net_name in nets]
             if not self._pedb.siwave.create_pin_group(self.reference_designator, pins, self.name):
-                raise ValueError(f"Failed to create pin group {self.name}")
+                raise RuntimeError(f"Failed to create pin group {self.name}")
         else:
             raise ValueError(f"No net and pins defined for defining pin group {self.name}")
 
--- a/src/pyedb/configuration/cfg_pin_groups.py
+++ b/src/pyedb/configuration/cfg_pin_groups.py
@@ -76,7 +76,7 @@ class CfgPinGroup(CfgBase):
             if not self._pedb.siwave.create_pin_group(self.reference_designator, pins, self.name):
                 raise RuntimeError(f"Failed to create pin group {self.name}")
         else:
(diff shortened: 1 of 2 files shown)


src/pyedb/configuration/cfg_pin_groups.py
```diff
@@ -67,17 +67,16 @@ def __init__(self, pedb, **kwargs):
...
-                comp = self._pedb.components.instances[self.reference_designator]
-            else:
-                raise Exception(f"Component not found for creating pin group {self.name}.")
-            pins = [p for p, obj in comp.pins.items() if obj.net_name in self.net]
+            nets = self.net if isinstance(self.net, list) else [self.net]
+            comp = self._pedb.components.instances[self.reference_designator]
+            pins = [p for p, obj in comp.pins.items() if obj.net_name in nets]
             if not self._pedb.siwave.create_pin_group(self.reference_designator, pins, self.name):
-                self._pedb.logger.error(f"Failed to create pin group {self.name}")
+                raise ValueError(f"Failed to create pin group {self.name}")
```
SMoraisAnsys, line 77:
```suggestion
                raise RuntimeError(f"Failed to create pin group {self.name}")
```

src/pyedb/configuration/cfg_pin_groups.py
```diff
@@ -67,17 +67,16 @@ def __init__(self, pedb, **kwargs):
...
         else:
-            self._pedb.logger.error(f"No net and pins defined for defining pin group {self.name}")
+            raise ValueError(f"No net and pins defined for defining pin group {self.name}")
```
SMoraisAnsys, line 79:
```suggestion
            raise RuntimeError(f"No net and pins defined for defining pin group {self.name}")
```

src/pyedb/dotnet/edb_core/siwave.py
```diff
@@ -1242,8 +1242,7 @@ def create_pin_group(self, reference_designator, pin_numbers, group_name=None):
         )
 
         if edb_pingroup.IsNull():  # pragma: no cover
-            self._logger.error(f"Failed to create pin group {group_name}.")
-            return False
+            raise ValueError(f"Failed to create pin group {group_name}.")
```
SMoraisAnsys, line 1245:
```suggestion
            raise RuntimeError(f"Failed to create pin group {group_name}.")
```

SMoraisAnsys (approved):
LGTM but I would raise `RuntimeError` instead.

Raising `ValueError` is dedicated to cases where """operation or function receives an argument that has the right type but an inappropriate value""". Here, that's not the case :)

SMoraisAnsys (approved):

PR #968, opened 2025-01-09
CI: Updates to use the new self-hosted runner

SMoraisAnsys:
For unknown reason the previous VM stopped working and no fix could be found, a new one was created and it uses standard path for AEDT installation.

I don't know how I succeeded to make all tests to work with the previous VM but, despite all my efforts, I'm no longer able to run test that leverage export_3d_cad aka export_to_hfss, export_to_q3d and export_to_maxwell. The reason is related to Siwave which must be called directly (through /opt/AnsysEM/v242/Linux64/siwave) leading to rendering the UI which is an issue in a non graphical environment.
Note that when using X virtual framebuffer, the tests are just taking too much time...

For the time being I've updated the PR to skip those 3 tests when they are run in linux on CICD. Note that you can still run them if you have a graphic card. For example, I'm able to run the test on export_to_hfss on Linux with my own laptop.

image

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -139,7 +139,7 @@ jobs:
     name: Test dotnet (linux)
     runs-on: [ Linux, self-hosted, toolkits ]
     env:
-      ANSYSEM_ROOT242: '/ansys_inc/AnsysEM/v242/Linux64'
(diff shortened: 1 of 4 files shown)


svandenb-dev (dismissed):
LGTM

hui-zhou-a (approved):

PR #977, opened 2025-01-14
CHORE: Bump dev version

svandenb-dev:
Bump release 0.36.0 dev0

### DIFF
--- a/src/pyedb/__init__.py
+++ b/src/pyedb/__init__.py
@@ -44,7 +44,7 @@ deprecation_warning()
 #
 
 pyedb_path = os.path.dirname(__file__)
-__version__ = "0.35.dev0"
+__version__ = "0.36.dev0"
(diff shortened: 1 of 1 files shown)


src/pyedb/__init__.py
```diff
@@ -44,7 +44,7 @@ def custom_show_warning(message, category, filename, lineno, file=None, line=Non
 #
 
 pyedb_path = os.path.dirname(__file__)
-__version__ = "0.35.dev0"
+__version__ = "0.36.dev0"
```
SMoraisAnsys, line 47:
```suggestion
__version__ = "0.37.dev0"
```

SMoraisAnsys (changes requested):
Seems like a new version has been released since then.
The version should be bumped to `0.37.dev0`.

Issue #988, opened 2025-01-18
CI: Failing to pass pyaedt's workflow test on linux


SMoraisAnsys:
There is something wrong with the latest release workflow as the latest run was not able to pass PyAEDT's workflow tests despit https://github.com/ansys/pyaedt/pull/5673, see the workflow log failure https://github.com/ansys/pyedb/actions/runs/12765093240/job/35787780022

The strangest thing is that I've been able to go on the VM and run the tests manually... See the result below.

![Image](https://github.com/user-attachments/assets/e2306245-ba6e-4b73-9837-c9cec514edec)

Note that I used an environment as close as possible to the one in CI by using the CI cached python3.10 interpreter and extending environment variables with `/home/ansys/actions-runner/_work/_tool/Python/3.10.16/x64/lib` to be able to use this interpreter.

svandenb-dev:
@SMoraisAnsys can we close this issue ?

SMoraisAnsys:
> [@SMoraisAnsys](https://github.com/SMoraisAnsys) can we close this issue ?

@svandenb-dev Thanks for the reminder, this PR can be closed yeah !

PR #987, opened 2025-01-17
CI: Temporary removal of PyAEDT dotnet tests

SMoraisAnsys:
Following https://github.com/ansys/pyedb/pull/986, this PR aims at avoiding the tests that are currently failing when releasing.

For visibility and future reference, here is a mini reproducer of the failure within the VM

```python
import os
from ansys.aedt.core.generic.grpc_plugin_dll_class import AEDT
os.environ["DesktopPluginPyAEDT"] = "/opt/AnsysEM/v242/Linux64/PythonFiles/DesktopPlugin"
AEDT(os.environ["DesktopPluginPyAEDT"])

from pythonnet import load
load(
    "coreclr",
    runtime_config=str("/usr/share/dotnet/sdk/6.0.428/dotnet.runtimeconfig.json"),
    dotnet_root=str("/usr/share/dotnet")
)
```

Note that removing the first 4 lines of the previous snippet allows to load `coreclr` correctly.

PR #990, opened 2025-01-21
CI: Deselect from project root

SMoraisAnsys:
Previous deselection (https://github.com/ansys/pyedb/pull/989) was failing because it must be specified from project root.
I tried it locally and from a folder `src` where I cloned `pyedb`, when I run `pytest pyedb/tests/legacy/unit --deselect=tests/legacy/unit/test_edb.py -vvv` the edb tests are correctly unselected (11 unselected tests)

For more information on that deselection requirement see https://github.com/ansys/pyedb/pull/987 and https://github.com/ansys/pyedb/issues/988
Note that https://github.com/ansys/pyedb/commit/da26422756f2dcf972ab659a4777354e7bf76251 was able to pass the file test that was previously failing.


svandenb-dev (approved):
LGTM

PR #992, opened 2025-01-23
FIX: merge via filtered by start and stop layers

boyang2022:
This pull request added padstack instances selection by start and stop layers in cases where vias belong to different layers but overlap in the same polygon.

Speed up the searching through self.instances.


svandenb-dev (approved):
LGTM

PR #994, opened 2025-01-23
FIX: Import gd s

gkorompi:


src/pyedb/dotnet/edb_core/edb_data/control_file.py
```diff
@@ -1262,6 +1265,42 @@ def parse_xml(self, xml_input):
...
+                            name=port_el.attrib["Name"],
+                            x1=port_el.attrib["x1"],
+                            y1=port_el.attrib["y1"],
+                            layer1=port_el.attrib["Layer1"],
+                            x2=port_el.attrib["x2"],
+                            y2=port_el.attrib["y2"],
+                            layer2=port_el.attrib["Layer2"],
+                            z0=port_el.attrib["Z0"],
+                        )
+            if el.tag == "SimulationSetups":
```
svandenb-dev, line 1281: Can you use find from ElementTree instead of these nested for loops ? 
This makes this hard to read and maintain

svandenb-dev (approved):
LGTM

PR #995, opened 2025-01-24
FIX: Get primitive from point bug fix with terminal

svandenb-dev:
Fixing a bug with method get_primitive_by_layer_and_point failing when a Terminal is return as the Class Terminal does not expose the GetPrimitiveType() method. 

closing #996


nefelimet (approved):

nefelimet (approved):
LGTM

PR #999, opened 2025-01-28
FEAT: Edb cfg equipotential

hui-zhou-a:
- Move _em_proerties to Primitive class, since it is accessible for all primitives not only padstack.
- EDB API has a bug in equipotential region. The workaround is to create a primitive and make it equipotential region
- New feature in edb config. The user is able to define an equipotential region for sources with point terminals.
- Deleted some redundant lines


svandenb-dev (approved):
LGTM

PR #1006, opened 2025-02-06
FIX: Edb cfg discrete

hui-zhou-a:


svandenb-dev (approved):
LGTM

Shih-Wei-Lin:
It seems like this modification doesn't work. Sorry, but it appears that in the aedb def file, you need to change IsDiscrete to True in order to make it discrete.

Shih-Wei-Lin:
When I use freq_sweep_type = "kDiscreteSweep", it doesn't change either.

Shih-Wei-Lin:
@hui-zhou-a  
Thank you for your response. However, PyEDB does not have any program or function that can modify the isDiscrete parameter. Is there any method to set a discrete sweep?

hui-zhou-a:
@Samuelopez-ansys do you know if sweep type can be set to discrete by pyaedt?

Samuelopez-ansys:
Yes it is possible once the EDB file is imported in HFSS 3D Layout

PR #1008, opened 2025-02-06
FIX: Ipc padstack instances fix

svandenb-dev:
This PR is fixing the potential bug in ipc exporter. Some padstack instances might be missing.

closing #1009


hui-zhou-a (approved):
LGTM

PR #1067, opened 2025-03-12
FIX: edb import coordiante circuilt port

hui-zhou-a:
Create the net is net from cfg file doesn't exist in edb


svandenb-dev (approved):
LGTM

PR #1075, opened 2025-03-17
FEAT: Add material from stackup

hui-zhou-a:
If materials is used in stackup but not defined, add it to edb with default values.

default conductor and dielectric material property values are defined in class Edb.Materials


src/pyedb/configuration/cfg_stackup.py
```diff
@@ -55,6 +55,35 @@ def __init__(self, pedb, data):
         self.materials = [CfgMaterial(**mat) for mat in data.get("materials", [])]
         self.layers = [CfgLayer(**lay) for lay in data.get("layers", [])]
 
+        materials = [m.name for m in self.materials]
+        for i in self.layers:
+            if i.type == "signal":
+                if i.material not in materials:
+                    self.materials.append(
+                        CfgMaterial(name=i.material, conductivity=580000000, permittivity=1, permeability=0.999991)
```
SMoraisAnsys, line 63: Just a question, do we have the `Material` and `Materials` classes aligned with the current default values ? Since they both can be used to create materials, we should ensure consistency within a common place / dictionary so that if one requires to be changed then the other is impacted in a consistent way.

src/pyedb/configuration/cfg_stackup.py
```diff
@@ -55,6 +55,35 @@ def __init__(self, pedb, data):
...
+                    )
+                if i.fill_material not in materials:
```
SMoraisAnsys, line 65: Doesn't those two if statement can lead to the creation of the same material twice ? This shouldn't be possible right ?

SMoraisAnsys (changes requested):
LGTM. Could you improve the consistency with the `Material` and `Materials` classes so that changes from one side also affects the other ? I left another minor comment

src/pyedb/configuration/cfg_stackup.py
hui-zhou-a: These values are copied from .amat from sys_library. They are not aligned with Material/Materials. I agree with you there should be a common place to hold the default material property values. I created an issue #1082

src/pyedb/configuration/cfg_stackup.py
hui-zhou-a: It is a bug. Fixed in the new comment.

SMoraisAnsys (dismissed):
LGTM - I would encourage to handle #1082 in this PR but, if it unlock an user workflows / scripts, it can be fixed in a second PR.
Note that this must be performed soon to avoid inconsistency between the creation approaches !

hui-zhou-a:
> LGTM - I would encourage to handle #1082 in this PR but, if it unlock an user workflows / scripts, it can be fixed in a second PR. Note that this must be performed soon to avoid inconsistency between the creation approaches !

Sure, I can implement in this PR.

src/pyedb/configuration/cfg_stackup.py
svandenb-dev: I will merge this PR and defer to issue #1082

PR #1077, opened 2025-03-17
FIX: cfg_padstacks.py

hui-zhou-a:


SMoraisAnsys (approved):

src/pyedb/configuration/cfg_padstacks.py
```diff
@@ -427,7 +427,7 @@ def __init__(self, pedb, pyedb_obj, **kwargs):
             self.api = self.DotNet(self)
 
         self.name = kwargs.get("name", None)
-        self.net_name = kwargs.get("net_name", "")
+        self.net_name = kwargs.get("net_name", None)
```
SMoraisAnsys, line 430: Can you add a test to avoid regressions later on ?

hui-zhou-a: I do have a use case, but this bug is due to a weird behavior of EDB API. I extended the PR description.

SMoraisAnsys (approved):

PR #1080, opened 2025-03-17
FEAT: edb cfg export wave port

hui-zhou-a:
Add support for exporting gap/wave port


hui-zhou-a:
> LGTM. Could you extend your PR description about the reason of the changes ?
> 
> I was going to request changes because I didn't see early on the renaming of `CfgWavePort` into `CfgEdgePort`. Here is the comment:
> 
> This is not clear. Why would we use CfgEdgePort if the type is "wave_port" ? Can't we handle a case where the type is "edge_port" instead ? If not, I think we should rename a few things to make them more understandable.

Description extended. CfgWavePort is confusing because its name is not consistant with EDB API. Thanks for the comments.

PR #1093, opened 2025-03-28
FIX: export solder_ball from IC component

hui-zhou-a:
Add import/export solder_ball property support for IC component


hui-zhou-a:
@amichel0205

svandenb-dev (approved):
LGTM

PR #1095, opened 2025-03-28
FIX: Circuit port creation on multiple pins

isaacwaldron:
Fix create_port_on_pins with multiple positive pins, an incorrect conditional was resulting in the creation of pin ports instead of pin group ports in this case.

### DIFF
--- a/tests/legacy/system/test_edb.py
+++ b/tests/legacy/system/test_edb.py
@@ -2002,6 +2002,21 @@ class TestClass:
         )
         assert len(edbapp.excitations) == 2
 
+    def test_create_circuit_port_on_component_pins_pingroup_on_multiple_pins(self, edb_examples):
+        edbapp = edb_examples.get_si_verse()
+        component_name = "U1"
+        edbcomp = edbapp.components[component_name]
+        positive_pin_names = ["R20", "R21", "T20"]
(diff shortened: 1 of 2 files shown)


tests/legacy/system/test_edb.py
```diff
@@ -2002,6 +2002,21 @@ def test_create_circuit_port_on_component_pins_pins_mode(self, edb_examples, pin
...
+        positive_pin_names = ["R20", "R21", "T20"]
+        reference_pin_names = ["N21", "R19", "T21"]
+        assert edbapp.components.create_port_on_pins(
+            refdes=edbcomp,
+            pins=positive_pin_names,
+            reference_pins=reference_pin_names,
+        )
+        assert len(edbapp.excitations) == 2
+        for excitation in edbapp.excitations.values():
+            assert excitation.terminal_type == "PinGroupTerminal"
```
isaacwaldron, line 2018: @svandenb-dev I am not sure if this check is compatible with gRPC, can you review this in particular?  Thank you!

svandenb-dev: The returned grpc will have a different type like no capital and "_" instead to be more pythonic. But .NET anf grpc are running the same tests but on different location so we will adjust accordingly.

PR #1114, opened 2025-04-07
FEAT: grpc configuratiom file boundaries

svandenb-dev:
This PR is adding grpc support for configuration file with boundaries.


hui-zhou-a (changes requested):
@svandenb-dev as we discussed a while ago with Massimo, please follow the same code structure as https://github.com/ansys/pyedb/blob/main/src/pyedb/configuration/cfg_components.py

Use subclass Common, Grpc, DotNet

maxcapodi78 (approved):

PR #1126, opened 2025-04-13
FIX: port location

hui-zhou-a:
- Move method location from Terminal class to PointTerminal  class
- Add new metho location to PadstackInstance class


svandenb-dev (approved):
LGTM

PR #1144, opened 2025-04-18
FIX: Cfg stackup refactoring

svandenb-dev:
This PR is refactoring the stackup cfg.


hui-zhou-a (commented):
@svandenb-dev grpc has a new stackup implementation as you mentioned before. Can you use grpc native method to import stackup?

PR #1163, opened 2025-04-25
FEAT: anti_pads_always_on and padstack rotation

hui-zhou-a:
- New property anti_pads_always_on to replace antipads_always_on
- padstacks.place accepts rotation as a string
- refactor edb cfg to support above two


svandenb-dev (approved):
LGTM

PR #1166, opened 2025-04-25
FIX: Grpc import failure fix

svandenb-dev:
This PR is fixing a wrong grpc import into .NET api.


Samuelopez-ansys (approved):

isaacwaldron (approved):
This was my mistake, oops!

PR #1176, opened 2025-05-02
FEAT: center line polygon data

hui-zhou-a:
- new property Edb.pyedb_class
- new method get_center_line_polygon_data
- new methods under PointData and PolygonData


svandenb-dev (approved):
LGTM

PR #1199, opened 2025-05-08
FIX: Fix unit tests that failed due to recent pyedb-core change in SweepData API

drewm102:
Fixes unit tests that are failing due to a recent pyedb-core change in SweepData API


svandenb-dev (approved):
LGTM

PR #1205, opened 2025-05-11
FIX: Edb wrapper method

hui-zhou-a:
Missing control_file arg

### DIFF
--- a/src/pyedb/generic/design_types.py
+++ b/src/pyedb/generic/design_types.py
@@ -33,6 +33,7 @@ def Edb(
     use_ppe=False,
     technology_file=None,
     grpc=False,
+    control_file=None,
 ):
     """Provides the EDB application interface.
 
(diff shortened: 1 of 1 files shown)


SMoraisAnsys (approved):
LGTM
@hui-zhou-a Was that argument removed in a new release ? Or, was it provided to `Edb` in another approach previously ? I would like to understand how the example started to fail and why this change is required.

hui-zhou-a:
> LGTM @hui-zhou-a Was that argument removed in a new release ? Or, was it provided to `Edb` in another approach previously ? I would like to understand how the example started to fail and why this change is required.

@SMoraisAnsys import_layout_file and Edb accept both control file and technology file. However, Edb always convert tech file into control file. I don't know why. The GDS import is unneccesariely complex. Edb wraper method -> Edb.__init__ -> import_layout_file. The problem was on Edb wrapper method.

PR #1220, opened 2025-05-16
FEAT: dot_net grpc compatibility

svandenb-dev:
This PR is adding the following:
- Bringing backward and forward compatibility between DotNet and grpc version
- Add a common (still duplicated test file for dotnet and grpc) having the exact same content to prepare the transition.
- The file is under test/system/test_edb,py which is the highest level test.
- The others files will come soon with step by step PR's

### DIFF
--- a/tests/grpc/system/test_edb.py
+++ b/tests/grpc/system/test_edb.py
@@ -255,6 +255,7 @@ class TestClass:
         points.append([bounding[0][0], bounding[0][1]])
 
         output = os.path.join(self.local_scratch.path, "cutout2.aedb")
+
         assert edbapp.cutout(
             custom_extent=points,
             signal_list=["GND", "1V0"],
@@ -285,7 +286,6 @@ class TestClass:
                 "LVDS_CH04_N",
             ],
             extent_type="Bounding",
-            number_of_threads=4,
             extent_defeature=0.001,
             preserve_components_with_model=True,
             keep_lines_as_path=True,
@@ -315,7 +315,6 @@ class TestClass:
         assert edbapp.cutout(
             signal_list=["1V0"],
             reference_list=["GND"],
-            number_of_threads=4,
             extent_type="ConvexHull",
             custom_extent=points,
             simple_pad_check=False,
@@ -338,7 +337,6 @@ class TestClass:
         assert edbapp.cutout(
             signal_list=["5V"],
             reference_list=["GND"],
-            number_of_threads=4,
             extent_type="ConvexHull",
             use_pyaedt_extent_computing=True,
             check_terminals=True,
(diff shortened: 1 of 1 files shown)


src/pyedb/dotnet/database/cell/terminal/terminal.py
```diff
@@ -195,6 +195,15 @@ def is_reference_terminal(self):
...
+    def reference_terminal(self):
+        """Adding grpc compatibility."""
+        return self.ref_terminal
+
+    @reference_terminal.setter
+    def reference_terminal(self, value):
+        self.ref_terminal = value
+
     @property
     def ref_terminal(self):
```
hui-zhou-a, line 208: can you add deprecation on this property as Massimo suggests?

src/pyedb/dotnet/database/sim_setup_data/data/mesh_operation.py
```diff
@@ -47,6 +47,15 @@ def __init__(self, parent, edb_object):
             "kNumMeshOpTypes": self._edb_object.TMeshOpType.kNumMeshOpTypes,
         }
 
+    @property
+    def net_layer_info(self):
+        """Adding property for grpc compatibility."""
```
hui-zhou-a, line 52: Can you also copy the docstring from grpc?

src/pyedb/grpc/database/components.py
```diff
@@ -276,20 +276,26 @@ def refresh_components(self):
...
-            elif i.type == "capacitor":
-                self._cap[i.name] = i
-            elif i.type == "inductor":
-                self._ind[i.name] = i
-            elif i.type == "ic":
-                self._ics[i.name] = i
-            elif i.type == "io":
-                self._ios[i.name] = i
-            elif i.type == "other":
+            try:
```
hui-zhou-a, line 279: Is it possible to avoid using try-except?

svandenb-dev: No sorry because the grpc internal api crashes. We spoke with @SMoraisAnsys about this and we will add a codacy exception to prevent warnings. I am in with dev to improve this behvior, that was actually done on design. While DotNet never crashes without throwing errors the grpc is the opposite. That's btw how I saw that DotNet tests were not tested properly and even failing silently.

src/pyedb/dotnet/database/sim_setup_data/data/mesh_operation.py
svandenb-dev: Done

src/pyedb/dotnet/database/cell/terminal/terminal.py
svandenb-dev: Done thanks for pointing this out.

hui-zhou-a (approved):

PR #1224, opened 2025-05-20
FEAT: grpc dotnet components compatibility

svandenb-dev:
This PR is adding compatibility between Components class including tests on grpc and dotnet version. 
grpc and dotnet have the exact same test_edb_components.py file.


gkorompi (approved):
Lgtm

PR #1235, opened 2025-05-20
FIX: Pin name

gkorompi:


svandenb-dev (commented):
Sorry I can't approve this PR I had the same discussion yesterday with @hui-zhou-a as I wanted to do the exact same changes to be compliant with grpc. grpc version is doing like your changes.

Issue #188, opened 2023-10-05
In edb Method set_solution_multi_frequencies() not working as expected


amichel0205:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

I try to add an HFSS setup in existing edb file.
Hfss setup is correctly added: medium1 = edb.create_hfss_setup("Medium1")
Hfss adaptive setup type is correctly set: medium1.adaptive_settings.adapt_type = "kMultiFrequencies"

When I try to add a list of frequencies using:
medium1.set_solution_multi_frequencies(frequencies=(10e9, 8e9), max_num_passes=15 ,max_delta_s=0.01)

only one frequency appears in the HFSS setup the last one in the list, 
max_num_passes is not updated

### Steps To Reproduce

I try to add an HFSS setup in existing edb file.
Hfss setup is correctly added: medium1 = edb.create_hfss_setup("Medium1")
Hfss adaptive setup type is correctly set: medium1.adaptive_settings.adapt_type = "kMultiFrequencies"

When I try to add a list of frequencies using:
medium1.set_solution_multi_frequencies(frequencies=(10e9, 8e9), max_num_passes=15 ,max_delta_s=0.01)

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.7

### Installed packages

use updated main from 05OCT2023

gkorompi:
Hey Alain, @amichel0205 fixed in https://github.com/ansys/pyedb/pull/1238

PR #1238, opened 2025-05-20
FIX: Multi freq

gkorompi:
Multi frequency setup fix, as it was including only one freq
closing #188


svandenb-dev (commented):
Please add a test

gkorompi:
> Please add a test

Added test for all cases (single,broadband,multi)

PR #1244, opened 2025-05-21
FEAT: grpc compatibility test edb layout

svandenb-dev:
This PR is adding grpc compatibility with test edb layout,


gkorompi (approved):
Lgtm

Issue #1231, opened 2025-05-20
dotnet component_res_value is returning a string


svandenb-dev:
### Description of the current limitations

check test_instances in test_edb_component.py
when calling component.res_value for instance we expect to get a float not a string. grpc is retuning a float.

PR #1254, opened 2025-05-22
FIX: Res value, Ind_value, Cap_calue

gkorompi:
fixing  Res value, Ind_value, Cap_calue to return float instead of string

closing #1231


svandenb-dev (approved):
LGTM as it didn't break any tests.
@hui-zhou-a be aware of this changes if you use it for your cfg file. 
I filled this issue as I saw ditnet version was returning a string on property calls res_value.
grpc is returning a float.
Thanks @gkorompi for taking care of this.

hui-zhou-a:
@svandenb-dev what if the cap value is a variable, how can I get the variable name?

Issue #1268, opened 2025-05-27
default adaptive frequency are not properly defined.


svandenb-dev:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

WE found the issue we must change from "1GHz" to "1Ghz"

### Steps To Reproduce

NA

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.7

### Installed packages

NA

PR #1269, opened 2025-05-27
FIX: adaptive frequency fix

svandenb-dev:
This PR is fixing EDB adaptive frequency value with changing default values from "1GHz" to "1Ghz". This was causing bad defintion in aedt. 

closing #1268


gkorompi:
why are smoke-tests failing? is it something negligible?

PR #1314, opened 2025-06-26
FIX: Update unit tests to be compatible with upcoming pyedb-core pre-release package

drewm102:
Update unit tests to be compatible with upcoming pyedb-core pre-release package v0.2.0.dev5

### DIFF
--- a/tests/grpc/system/test_edb_stackup.py
+++ b/tests/grpc/system/test_edb_stackup.py
@@ -800,8 +800,8 @@ class TestClass:
                 assert cellInstance.Is3DPlacement()
             if chipEdb.grpc:
                 transform = cellInstance.transform3d
-                assert transform.matrix[3][:3] == [0, 0, 0.00016]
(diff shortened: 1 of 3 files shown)


tests/grpc/system/test_edb_stackup.py
```diff
@@ -800,8 +800,8 @@ def test_stackup_place_on_top_of_lam_with_mold_solder(self):
                 assert cellInstance.Is3DPlacement()
             if chipEdb.grpc:
                 transform = cellInstance.transform3d
```
drewm102, line 802: Soldberball properties for `ComponentType.OTHER` were fixed in [this PR](https://github.com/ansys/pyedb-core/pull/562). These unit tests had to be updated to reflect these changes (let me know if this doesn't seem right to you though)

svandenb-dev: I don't see Solderballs involved in this test. Should it be another ?

PR #1333, opened 2025-07-04
REFACTOR: design_types.py

hui-zhou-a:
This PR makes available methods/properties show up in IDE

image


src/pyedb/generic/design_types.py
```diff
@@ -34,7 +38,7 @@ def Edb(
     technology_file=None,
     grpc=False,
     control_file=None,
-):
+) -> EdbDotnet:
```
SMoraisAnsys, line 41: This isnt true as EDBGrpc can also be returned. Please update the type hint

PR #1348, opened 2025-07-08
FEAT: DC settings

gkorompi:


gkorompi:
> dc_advanced_settings It is already in there.
> 
> https://github.com/ansys/pyedb/blob/c6ebb4a24e00bf06df02cc04d3070d65704d5a9a/src/pyedb/dotnet/database/utilities/siwave_simulation_setup.py#L365

Hello @hui-zhou-a i think this is in SiwaveDCSimulationSetup class, we need it in SiwaveSimulationSetup too

hui-zhou-a:
@svandenb-dev @gkorompi revert to GK's commit as changing code structure is not that simple. Lots of test are failing.

svandenb-dev (approved):
LGTM

svandenb-dev (approved):
LGTM

PR #1410, opened 2025-07-24
FEAT: Import stackup from vlc.tech

gkorompi:


src/pyedb/grpc/edb.py
```diff
@@ -838,6 +838,53 @@ def import_layout_file(
...
+
+        Returns
+        -------
+        Full path to the AEDB file : str
+
+        """
+        if not working_dir:
+            working_dir = os.path.dirname(vlctech_file)
+        command = os.path.join(self.base_path, "helic", "tools", "raptorh", "bin", "make-edb")
+        if is_windows:
```
svandenb-dev, line 867: How about Linux ? you can do like you did for edb_diff ?

svandenb-dev (commented):
Please add Linux support otherwise great job!!

src/pyedb/grpc/edb.py
gkorompi: thanks simon

svandenb-dev (approved):
LGTM

PR #1411, opened 2025-07-24
FEAT: Design libraries

svandenb-dev:
This PR is adding new RF libraries


svandenb-dev:
> @svandenb-dev would you like to add example/user guide?

I made example in doc string and also added a new tab in online menu. Where we will put all libraries

Issue #1399, opened 2025-07-23
Base path error in linux with 25R2 when using ViaDesignBackend


SMoraisAnsys:
One of PyAEDT's test is failing in Linux with 25R2 because it is interacting with `ViaDesignBackend`.
I've created a reproducer that doesn't require interaction with the extension:

```python
    from pathlib import Path

    from pyedb.extensions.via_design_backend import ViaDesignBackend

    from ansys.aedt.core.extensions.misc import get_aedt_version
    from ansys.aedt.core.extensions.misc import get_port
    from ansys.aedt.core.extensions.misc import get_process_id
    from ansys.aedt.core.extensions.misc import is_student
    from ansys.aedt.core.hfss3dlayout import Hfss3dLayout

    PORT = get_port()
    VERSION = get_aedt_version()
    AEDT_PROCESS_ID = get_process_id()
    IS_STUDENT = is_student()

    def test_custom():
        # This needs to be updated to match your file location
        create_design_path = Path("use/pyaedt/project/resources/via_design_pcb_rf.toml")
        assert create_design_path.exists()
    
        dict_config = toml.load(create_design_path)
        stacked_vias = dict_config.pop("stacked_vias")
        for param_name, param_value in dict_config["signals"].items():
            stacked_vias_name = param_value["stacked_vias"]
            dict_config["signals"][param_name]["stacked_vias"] = stacked_vias[stacked_vias_name]
        for param_name, param_value in dict_config["differential_signals"].items():
            stacked_vias_name = param_value["stacked_vias"]
            dict_config["differential_signals"][param_name]["stacked_vias"] = stacked_vias[stacked_vias_name]
    
        # The following code line should fail
        backend = ViaDesignBackend(dict_config)
        print("Created via design backend.")
  ```

Here is the stacktrace I'm having on my machine

```shell
  >       backend = ViaDesignBackend(dict_config)
  
  external/pyaedt/tests/system/extensions/test_via_design.py:104: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  .venv/lib/python3.10/site-packages/pyedb/extensions/via_design_backend.py:676: in __init__
      self.app = Edb(
  .venv/lib/python3.10/site-packages/pyedb/generic/design_types.py:280: in Edb
      return app(
[...]

PR #1435, opened 2025-08-04
REFACTOR: Check installed versions

hui-zhou-a:
This is to fix https://github.com/ansys/pyedb/issues/1399


hui-zhou-a:
This is to fix #1399

Samuelopez-ansys (approved):

PR #1470, opened 2025-08-28
FEAT: Duplicate cell array

svandenb-dev:


hui-zhou-a (dismissed):
This is method is only for a special use case. I don't think it fits in modeler. Please move to extension folder.

PR #1476, opened 2025-09-01
BUILD(ci): bump ansys/actions into v10.0.15

SMoraisAnsys:
As title says. This will avoid `safety`'s dependence issue, fixed in https://github.com/ansys/actions/pull/971


svandenb-dev (approved):
LGTM

PR #1482, opened 2025-09-02
REFACTOR: Raise errors instead of using ``assert`` in code outside of tests

ecoussoux-ansys:
This PR provides refactoring for raising errors instead of using ``assert`` in code. Use of ``assert`` in test files is not regarded as problematic and is left unchanged. The modifications introduced remove all instances of ``assert`` in code outside from test files throughout the entire ``pyedb`` project.
The motivation for this update can be read from [here](https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html): In some cases, ``assert`` clauses may be removed upon code compilation, resulting in the loss of the protections implemented. Therefore, raising semantically meaningful errors should be preferred.

For these changes, similar updates previously performed in ``pyaedt`` have been leveraged as follows:
* ``src/pyedb/dotnet/database/Variables.py`` was updated following changes made to the ``pyaedt`` file ``src/ansys/aedt/core/application/variables.py`` in https://github.com/ansys/pyaedt/pull/4554 also considering updates later introduced to that same file in https://github.com/ansys/pyaedt/pull/5272,
* ``src/pyedb/generic/data_handlers.py`` was updated following changes made to the ``pyaedt`` file ``src/ansys/aedt/core/generic/data_handlers.py`` in https://github.com/ansys/pyaedt/pull/5995,
* ``src/pyedb/modeler/geometry_operators.py`` was updated following changes made to the ``pyaedt`` file ``src/ansys/aedt/core/modeler/geometry_operators.py`` in https://github.com/ansys/pyaedt/pull/5579,
* ``src/pyedb/siwave.py`` was updated following implementation of similar checks on version in the ``pyaedt`` files ``src/ansys/aedt/core/desktop.py`` and ``src/ansys/aedt/core/filtersolutions_core/dll_interface.py``.

### DIFF
--- a/src/pyedb/dotnet/database/Variables.py
+++ b/src/pyedb/dotnet/database/Variables.py
@@ -181,13 +181,15 @@ class CSVDataset:
                 if variable in key_string:
                     found_variable = True
                     break
-            assert found_variable, "Input string {} is not a key of the data dictionary.".format(variable)
+            if not found_variable:
+                raise KeyError(f"Input string {variable} is not a key of the data dictionary.")
             data_out._data[variable] = self._data[key_string]
             data_out._header.append(variable)
         return data_out
 
     def __add__(self, other):  # pragma: no cover
-        assert self.number_of_columns == other.number_of_columns, "Inconsistent number of columns"
+        if self.number_of_columns != other.number_of_columns:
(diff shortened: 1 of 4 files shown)


svandenb-dev (approved):
LGTM

PR #1530, opened 2025-09-16
REFACTOR: Improve security around use of ``subprocess``

ecoussoux-ansys:
This PR aims at improving security around the use of ``subprocess`` for executing commands inside scripts of ``pyedb``.
The motivation for the changes can be read from these sources, discussing the security implications associated with the ``subprocess`` module: 
* On the [import of ``subprocess``](https://bandit.readthedocs.io/en/latest/blacklists/blacklist_imports.html#b404-import-subprocess),
* On the [use of ``subprocess`` with shell equals true](https://bandit.readthedocs.io/en/latest/plugins/b602_subprocess_popen_with_shell_equals_true.html)
* On the [use of ``subprocess`` without shell equals true](https://bandit.readthedocs.io/en/latest/plugins/b603_subprocess_without_shell_equals_true.html)
* On [starting processes with a shell](https://bandit.readthedocs.io/en/latest/plugins/b605_start_process_with_a_shell.html)

Overall, invoking an external executable or executing a command through ``subprocess`` "may present a security issue if appropriate care is not taken to sanitize any user provided or variable input", potentially resulting in some cases to "vulnerability to various shell injection attacks".

The approach taken here is mainly derived from the work performed in https://github.com/ansys/pyaedt/pull/5892.
In short, each occurrence of ``subprocess.run`` is now contained in a ``try / except`` structure and a warning has been inserted in the docstrings of all the methods exhibiting such calls: This warning is linked to a new section of the user guide that has been added to summarize the security considerations associated with the use of ``subprocess`` in ``pyedb``.
On top of that, the argument ``check=True`` is systematically passed in calls to ``subprocess.run()`` to allow for exception handling as described [here](https://docs.python.org/3/library/subprocess.html): "If check is true, and the process exits with a non-zero exit code, a ``CalledProcessError`` exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured."

For further reference, this topic was also addressed on ``pyaedt`` in the following PRs: https://github.com/ansys/pyaedt/pull/5579, https://github.com/ansys/pyaedt/pull/6001, https://github.com/ansys/pyaedt/pull/6110.

### DIFF
--- a/src/pyedb/dotnet/database/edb_data/control_file.py
+++ b/src/pyedb/dotnet/database/edb_data/control_file.py
@@ -27,6 +27,8 @@ import subprocess  # nosec B404
 import sys
 import xml
 
+from defusedxml.ElementTree import parse as defused_parse
+
 from pyedb.generic.general_methods import ET, env_path, env_value, is_linux
 from pyedb.generic.settings import settings
 from pyedb.misc.aedtlib_personalib_install import write_pretty_xml
@@ -1219,7 +1221,7 @@ class ControlFile:
         -------
         bool
         """
-        tree = ET.parse(xml_input)
+        tree = defused_parse(xml_input)
         root = tree.getroot()
         for el in root:
             if el.tag == "Stackup":
--- a/src/pyedb/generic/process.py
+++ b/src/pyedb/generic/process.py
@@ -23,6 +23,54 @@ class SiwaveSolve(object):
         full_path = Path(self._pedb.ansys_em_path) / executable
         return str(full_path)
 
+    def __create_exec(self, type):
+        base = os.path.splitext(self._pedb.edbpath)[0]
+        txt_path = base + ".txt"
+        exec_path = base + ".exec"
+        with open(txt_path, "w") as file:
+            if type:
+                if type == "DCIR":
+                    file.write("ExecDcSim")
+                elif type == "SYZ":
+                    file.write("ExecSyzSim")
+                elif type == "CPA":
+                    file.write("ExecSentinelCpaSim")
+                elif type == "TimeCrosstalk":
+                    file.write("ExecTimeDomainCrosstalkSim")
+                elif type == "FreqCrosstalk":
+                    file.write("ExecCrosstalkSim")
+                elif type == "Impedance":
+                    file.write("ExecZ0Sim")
+
+        os.rename(txt_path, exec_path)
+        return exec_path
+
+    def solve_siwave(self, edbpath, analysis_type):
+        """Solve an SIWave setup. Only non-graphical batch mode is supported.
+
+        Parameters
+        ----------
+        analysis_type: str
+            Type of SIWave analysis to perform. Available types are "SYZ", "DCIR", "CPA", "TimeCrosstalk",
+            "FreqCrosstalk", "Impedance".
+        edbpath: str
+            Full path to the .aedb folder, siw or siwz file to be solved.
+        siwave_ng: str, optinial
+            Path to the siwave_ng. Default is the SIWave installation path.
+        """
+
+        command = [
+            self.__siwave_ng_exe_path,
+            edbpath,
+            self.__create_exec(type=analysis_type),
+            "-formatOutput",
+            "-useSubdir",
+        ]
+        try:
+            subprocess.run(command, check=True)
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError(f"An error occurred when launching the solver. Please check input paths") from e
+
     def solve(self, num_of_cores=4):
         """Solve using siwave_ng.exe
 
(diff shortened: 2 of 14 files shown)


src/pyedb/dotnet/database/edb_data/control_file.py
```diff
@@ -101,8 +106,10 @@ def convert_technology_file(tech_file, edbversion=None, control_file=None):
         commands.append(["rm", "-r", vlc_file_name + ".aedb"])
         my_env = os.environ.copy()
         for command in commands:
-            p = subprocess.Popen(command, env=my_env)
-            p.wait()
+            try:
+                subprocess.run(command, env=my_env, check=True)  # nosec
```
SMoraisAnsys, line 110: Are we sure about the need to pass `env=my_env` ? AFAIK the subprocess should start with the environment variables as the context run the command. Since `my_env` is just a copy of `os.environ` and not extended, I don't see why it's needed here.

src/pyedb/grpc/database/control_file.py
```diff
@@ -214,8 +219,10 @@ def convert_technology_file(tech_file, edbversion=None, control_file=None):
         commands.append(["rm", "-r", vlc_file_name + ".aedb"])
         my_env = os.environ.copy()
```
SMoraisAnsys, line 220: Same comment as before

src/pyedb/grpc/database/utility/xml_control_file.py
```diff
@@ -100,8 +105,10 @@ def convert_technology_file(tech_file, edbversion=None, control_file=None):
         commands.append(["rm", "-r", vlc_file_name + ".aedb"])
         my_env = os.environ.copy()
```
SMoraisAnsys, line 106: See all occurences

src/pyedb/misc/downloads.py
```diff
@@ -65,7 +73,10 @@ def _retrieve_file(url, filename, directory, destination=None, local_paths=[]):
     # Perform download
     if is_linux:
         command = "wget {} -O {}".format(url, local_path)
```
SMoraisAnsys, line 75: Here command should be a list and not a string.
Could you ensure that's also the case in every modification (I think I didn't find another string value but better check it) ?
Also, is this command only retrieve data from our examples ? If yes, could you add a check to ensure the the URL is a know one (e.g. pointing to ansys/example-data or something like that)

SMoraisAnsys (commented):
Thanks for the changes @ecoussoux-ansys ! I left some comments :)
On top of the comments, could you also cleanup the code so that variable names around subprocess calls make sense ? I've often seen things like `command = "i'm a string"`, and then defining `cmd_input = [command, "and this is another string"]` when `command` was the path to some EXE or other kind of file.

svandenb-dev (approved):
Thanks @ecoussoux-ansys  for your work. Adding @gkorompi  as he wrote some of these methods for information.

svandenb-dev (approved):
LGTM

src/pyedb/dotnet/edb.py
```diff
@@ -4852,16 +4865,16 @@ def compare(self, input_file, results=""):
...
-        if p.returncode == 0:
+            command = [executable_path, input_file, self.edbpath, results]
+        try:
+            subprocess.run(command, check=True)  # nosec
             return str(Path(self.base_path).joinpath("EDBDiff.exe"))
-        else:
+        except subprocess.CalledProcessError as e:  # nosec
             raise RuntimeError(
                 "EDBDiff.exe execution failed. Please check if the executable is present in the base path."
             )
```
SMoraisAnsys, line 4880:
```suggestion
            ) from e
```

src/pyedb/generic/process.py
```diff
@@ -97,9 +110,9 @@ def export_3d_cad(
         command += ["-RunScriptAndExit", scriptname]
         print(command)
         try:
-            result = subprocess.run(command, check=True, capture_output=True)
+            result = subprocess.run(command, check=True, capture_output=True)  # nosec
             print(result.stdout.decode())
-        except subprocess.CalledProcessError as e:
+        except subprocess.CalledProcessError as e:  # nosec
             print(f"Error occurred: {e.stderr.decode()}")
```
SMoraisAnsys, line 116: Shouldn't an exception be raised here ?

SMoraisAnsys (commented):
Just a last batch of comment, thanks for the changes @ecoussoux-ansys !!

src/pyedb/generic/process.py
ecoussoux-ansys: Indeed, I overlooked this one as I did not apply changes to it. 
Thanks for pointing out, I applied something in the line of the other occurrences in 8387add.

Issue #1507, opened 2025-09-10
Bug located in GRPC boundary_type


hui-zhou-a:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

<img width="816" height="265" alt="Image" src="https://github.com/user-attachments/assets/108fdcd0-f7d7-43e9-a607-a02f7f2b06ff" />

### Steps To Reproduce

```
edbapp = edb_examples.get_si_verse()
edbapp.hfss.create_coax_port_on_component("U1", "DDR4_DQS0_P")
edbapp.terminals["U1_DDR4_DQS0_P_E1"].boundary_type = "rlc"
```

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.7

### Installed packages

all

PR #1531, opened 2025-09-17
FIX: issue 1507 fix

svandenb-dev:
Fixing issue 1507
closing #1507


hui-zhou-a (changes requested):
Can you add tests?
image

PR #1543, opened 2025-09-19
REFACTOR: Simplify optional-dependencies and remove ``try-except`` patterns around imports from base dependencies

ecoussoux-ansys:
This PR introduces two main changes as follows:
* Simplifying the project's optional-dependencies (in ``pyproject.toml``) by removing the ``[full]`` target: From the 2 packages that were part of this target, ``shapely`` is already included in all other targets as well as in the base dependencies - ``matplotlib`` is moved to the project's base dependencies as it is in fact necessary for many operations (as was evidenced by messages/logs issued if it is not installed, for instance in files like ``src/pyedb/dotnet/database/stackup.py`` or ``src/pyedb/generic/plot.py``),

* Removing ``try-except`` patterns around imports of packages which belong to the project's base dependencies, as well as in-file checks of the form ``if not "package": log() / return False``. This was notably observed for some imports of ``pandas`` and ``numpy`` and was extended to such cases affecting ``matplotlib`` imports too, as the package was added to the project's base dependencies.

### DIFF
--- a/src/pyedb/dotnet/database/cell/hierarchy/component.py
+++ b/src/pyedb/dotnet/database/cell/hierarchy/component.py
@@ -25,6 +25,8 @@ import re
 from typing import Optional
 import warnings
 
+import numpy as np
+
 from pyedb.dotnet.database.cell.hierarchy.hierarchy_obj import Group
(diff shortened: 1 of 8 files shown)


svandenb-dev (approved):
LGTM

PR #1586, opened 2025-10-20
REFACTOR: Ipc update

gkorompi:


svandenb-dev:
@gkorompi if we want to keep backward compatibility for some time we should keep the files and check which version is used. If the one used is eligible to the Ansys translator one then use it if not fall back to previous export.

svandenb-dev (approved):
LGTM

PR #1607, opened 2025-10-25
FEAT: Design mode

gkorompi:


gkorompi:
in dotenet its named model_type. In grpc is exposed as design_mode. How should i name it?

svandenb-dev (approved):
LGTM

PR #1651, opened 2025-11-13
CI: Leverage new ``vtk-osmesa`` logic in CI

ecoussoux-ansys:
This PR is porting to ``pyedb`` the changes introduced in https://github.com/ansys/pyaedt/pull/6868.

In the CI, the installation of ``vtk-osmesa`` in the virtual environments created for the windows and linux ``pyaedt`` tests are dropped since the wheels that were used are no longer available. We are now using the [``pyvista/setup-headless-display-action``](https://github.com/pyvista/setup-headless-display-action) for these jobs instead.

Some other changes to the ``ci_cd.yml`` file introduced to ``pyaedt`` in https://github.com/ansys/pyaedt/pull/6868 are also reproduced.

### DIFF
--- a/doc/changelog.d/1651.maintenance.md
+++ b/doc/changelog.d/1651.maintenance.md
@@ -1,0 +1,1 @@
+Leverage new \`\`vtk-osmesa\`\` logic in CI


svandenb-dev (approved):
LGTM

PR #1661, opened 2025-11-21
CHORE: Delete accidentally added files

SMoraisAnsys:
Seems like a lot of non necessary files got added in #1581 Please have a look in the associated PR and ensure that I didn't miss other non needed files.


svandenb-dev (dismissed):
LGTM

PR #1694, opened 2025-12-03
REFACTOR: Random failure test

hui-zhou-a:


src/pyedb/__init__.py
```diff
@@ -59,7 +59,7 @@ def custom_show_warning(message, category, filename, lineno, file=None, line=Non
 #
 
 pyedb_path = os.path.dirname(__file__)
-__version__ = "0.66.dev0"
+__version__ = "0.66.dev00"
```
svandenb-dev: why this change ?

PR #1729, opened 2026-01-05
REFACTOR: edb grpc main class warning fix

svandenb-dev:


src/pyedb/grpc/edb.py
```diff
@@ -1430,7 +1471,7 @@ def import_cadence_file(self, inputBrd, WorkDir=None, anstranslator_full_path=""
 
     def import_gds_file(
         self,
-        inputGDS,
+        input_gds,
```
hui-zhou-a: Will it not break backward compatibility?

src/pyedb/grpc/edb.py
```diff
@@ -540,7 +550,7 @@ def ports(self) -> Dict[str, GapPort]:
         dict[str, list[:class:`GapPort` or :class:`WavePort` or :class:`CoaxPort`]]
             Port names and objects.
         """
-        terminals = [term for term in self.layout.terminals if not term.is_reference_terminal]
+        terminals = [term for term in self.layout.terminals if not getattr(term, "is_reference_terminal", False)]
```
hui-zhou-a: In which case is_reference_terminal doesn't exist?

src/pyedb/grpc/edb.py
svandenb-dev: Oh right. I added the decorator to prevent this break.

PR #1755, opened 2026-01-14
REFACTOR: edb_object replaced by core

svandenb-dev:
this PR is replacing all edb_object by core for grpc to be consistent with other classes and pyedb-core


hui-zhou-a (approved):

hui-zhou-a (dismissed):
Please add tests

PR #1804, opened 2026-02-03
FIX: Raptor fix

gkorompi:


hui-zhou-a (dismissed):
@gkorompi thanks for fixing this.  could you add tests?

PR #1828, opened 2026-02-14
REFACTOR: Lazy import and dependency rework

SMoraisAnsys:
Rework of the pyedb installation to avoid installing too many dependencies when some of them are only for a specific subset of features. This will help with the creation of a pyedb recipe for conda and also avoid downstream applications to install too many unneeded dependencies.


svandenb-dev (approved):
LGTM

PR #1836, opened 2026-02-16
CHORE: Clean up root level files

SMoraisAnsys:
As title says. Removal of some files not needed any more and update of others.
This should also help fixing the missing pre-commit updates.


svandenb-dev (approved):
LGTM

PR #1841, opened 2026-02-19
FEAT: Add fill material support to backdrill methods

anskprasad:
* PadstackInstance  now support an optional fill_material parameter for backdrill operations. 
* The set_back_drill_by_depth and set_back_drill_by_layer methods accept this argument, and the corresponding get methods return named tuples including fill material information. 
* Adding a new optional variable include_fill_material with default set to false, so that the users dont see a breakage in their existing scripts. 
* But to get fill material, the variable must be turned on. 
* Docstrings and type annotations are updated for clarity and improved API usability.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -738,7 +738,7 @@ jobs:
       - name: Install PyAEDT main branch version with its test dependencies
         run: |
           . .venv\Scripts\Activate.ps1
-          pip install --no-cache-dir external/pyaedt[tests]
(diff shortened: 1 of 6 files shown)


hui-zhou-a:
@anskprasad thanks for the new arguments. Which AEDT version starts to support fill material in backdrill?

anskprasad:
> @anskprasad thanks for the new arguments. Which AEDT version starts to support fill material in backdrill?

@hui-zhou-a It's targeted for 2027.1.

PR #1870, opened 2026-03-02
FIX: Fixed 2 tests in grpc

maxcapodi78:
fixed test in layout validation and fixed test on layers consolidated


svandenb-dev (dismissed):
LGTM

PR #1873, opened 2026-03-02
FIX: Cfg tests 2026.1 fix

svandenb-dev:
This PR is fixing 2 cfg tests. They are passing only with Release 2026.1 and pyedb-core dev7 version.

### DIFF
--- a/src/pyedb/grpc/database/components.py
+++ b/src/pyedb/grpc/database/components.py
@@ -965,7 +965,7 @@ class Components(object):
                 else:
                     new_cmp.component_type = CoreComponentType.CAPACITOR  # RLC type by default
                 pin_pair = (pins[0].name, pins[1].name)
-                rlc_model = PinPairModel(new_cmp.component_property.model)
+                rlc_model = PinPairModel(self._pedb, new_cmp.component_property.model)
                 rlc_model.core.set_rlc(pin_pair, rlc)
                 component_property = new_cmp.component_property
                 component_property.model = rlc_model.core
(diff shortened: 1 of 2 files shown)


src/pyedb/grpc/database/hierarchy/pin_pair_model.py
```diff
@@ -21,22 +21,66 @@
...
+        if not pin2_name:
+            pin2_name = "2"
+        core.set_rlc(pin_pair=(pin1_name, pin2_name), rlc=rlc)
+        return cls(core)
 
     @property
     def first_pin(self) -> str:
-        """First pin name.
+        """First pin name. This attribute is read-only since pin pair model is defined between two pins,
+        and changing pin names will change the pin pair itself.
```
SMoraisAnsys:
```suggestion
        """First pin name.
        
        This attribute is read-only since pin pair model is defined between two pins,
        and changing pin names will change the pin pair itself.
```

src/pyedb/grpc/database/hierarchy/pin_pair_model.py
```diff
@@ -49,7 +93,8 @@ def first_pin(self) -> str:
 
     @property
     def second_pin(self) -> str:
-        """Second pin name.
+        """Second pin name. This attribute is read-only since pin pair model is defined between two pins,
+        and changing pin names will change the pin pair itself.
```
SMoraisAnsys:
```suggestion
        """Second pin name.
        
        This attribute is read-only since pin pair model is defined between two pins,
        and changing pin names will change the pin pair itself.
```

hui-zhou-a (approved):

PR #1874, opened 2026-03-02
FIX: Hfss simulation setup grpc

svandenb-dev:
This PR is fixing grpc hfss simulation setup test with keeping DotNet compatilibility. It also introduces the parent class inside Options classes to allow navigation across all settings since DotNet has different structure. Deprecation has also been added to move on new grpc architecture in the future.


Samuelopez-ansys:
@svandenb-dev please answer @SMoraisAnsys comment

PR #1878, opened 2026-03-02
FIX: improve docstring and coverage for .net

gkorompi:


src/pyedb/dotnet/database/cell/hierarchy/hierarchy_obj.py
```diff
@@ -30,13 +31,25 @@ def __init__(self, pedb, edb_object):
...
     @property
-    def location(self):
-        """XY Coordinates."""
+    def location(self) -> List | None:
+        """Return XY coordinates if available.
+
+        Returns
+        -------
+        list or None
+            [x, y] if available, else None.
```
svandenb-dev:
```suggestion
            (x, y) if available, else None.
```

SMoraisAnsys (changes requested):

PR #1895, opened 2026-03-10
CHORE: Deprecation decorator

svandenb-dev:
This PR is replacing deprecation warning using decorator for grpc and DotNet when applicable.


Samuelopez-ansys (dismissed):

Samuelopez-ansys (dismissed):

SMoraisAnsys (commented):
I'm not in favor of adding automation about inferring replacement from a return statement. I'm pretty sure there will be caveat when using it and this probably won't be found by us but users. Why not just providing the name of the method to use as an argument of the functions ?

hui-zhou-a (commented):
@svandenb-dev can you add deprecation information in docstring also?

hui-zhou-a (dismissed):
The IDE inspector doesn't recognize the deprecation 

Image

This is before 

Image

maxcapodi78 (dismissed):

maxcapodi78 (approved):

PR #1910, opened 2026-03-12
REFACTOR: remove unused method to retrieve reference primitive from terminal

### LINKED ISSUE #1903: Check get_edge_terminal_reference_primitive for grpc

### Description of the bug

This method does not work (most likley)

### Steps To Reproduce

NA

Windows

3.10

### Installed packages

NA

svandenb-dev:
This PR is removing unused method to retrieve reference primitive from terminal and tries instead using pyedb-core api to get it when available. The removed code was not working. Instead raising an error message telling what to use, as it is not a deprecation but rather another approach. In any case the previous one could not work.

### DIFF
--- a/doc/changelog.d/1910.miscellaneous.md
+++ b/doc/changelog.d/1910.miscellaneous.md
@@ -1,0 +1,1 @@
+Remove unused method to retrieve reference primitive from terminal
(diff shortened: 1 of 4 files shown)


ecoussoux-ansys (commented):
I am not able to assess the logic implemented here to replace the old, dysfunctional method, sorry.

Just two comments : 
* Is that same method working fine in dotnet? That one is currently not covered by tests. (Sorry if it's not a good practice to mention dotnet in a grpc PR.. I’m not sure whether you usually deal with the two separately, or whether an update to a method present in both is seen as a good opportunity to review the counterpart)
* The new logic is currently not completely covered by tests, maybe this could be improved?

Image

maxcapodi78 (approved):

PR #1985, opened 2026-04-02
FIX: stackup.py roughness import

hui-zhou-a:


svandenb-dev (commented):
I see you are adding line in xml file for roughness and etching but there is no test code associated. Can you also make sure it's working both on grpc and dotnet ?

PR #2034, opened 2026-04-13
FIX: improve grpc Nets.delete performance for large net deletions

### LINKED ISSUE #2035: grpc Nets.delete is extremely slow for large net deletions and emits null-net errors
## Description
`pyedb.grpc.database.nets.Nets.delete` becomes extremely slow when deleting a large set of nets and can emit repeated null-net errors during the operation.

## Repro shape
On a real design, deleting 1911 nets from an initial 2158-net layout through `Nets.delete(...)` produced:
- very long runtime
- repeated `InvalidArgumentException('Illegal operation on a null Net object')` messages
- blank entries in the deleted-net sample output

The final layout state was still correct, but the path was much slower and noisier than expected.

## Benchmark
Using the same scenario and layout for both implementations:
- initial net count: `2158`
- requested deletions: `1911`
- final net count: `247`

Legacy `Nets.delete`:
- `delete_s=1576.161`

Updated net-centric traversal:
- `delete_s=26.262`

That is about a `60x` difference.

## Suspected cause
The current implementation performs full-layout scans through:
- `modeler.delete_primitives(...)`
- `padstacks.delete_padstack_instances(...)`
- then re-enumerates nets for the final delete pass

This appears to create a large amount of unnecessary work for big delete sets and may also contribute to invalid/null net wrapper access during the tail end of the operation.

## Proposed fix
A PR is available that changes `Nets.delete` to:
- traverse the requested net objects directly
- collect attached primitives and padstack instances from those nets only
- preserve the existing delete order: `primitives -> padstacks -> nets`

PR: #2034

## Validation
- exact keep-set verification passed after deletion
- the resulting `.aedb` was opened and inspected in SIWave and the deleted nets were removed as expected

pl170491:
## Summary
- rework `grpc.database.nets.Nets.delete` to traverse the requested net objects directly instead of scanning all layout primitives and padstack instances first
- preserve the existing delete order (`primitives -> padstacks -> nets`) and overall behavior while avoiding null-net churn during large delete operations
- add a focused system test covering multi-net deletion with nonexistent requested net names

## Why
`Nets.delete` currently scales poorly for large delete sets because it performs full-layout scans through `modeler.delete_primitives(...)` and `padstacks.delete_padstack_instances(...)` before deleting the requested nets themselves.

On a real design where 1911 nets were deleted from an initial 2158-net layout, the current implementation took:
- legacy: `1576.161s`
- updated implementation: `26.262s`

This is about a `60x` speedup.

In the same benchmark, the legacy path also emitted a large number of:
- `InvalidArgumentException('Illegal operation on a null Net object')`

and returned blank entries in the deleted-net sample, while the updated path completed cleanly.

## Validation
Using the same scenario and layout for both runs:
- initial net count: `2158`
- requested deletions: `1911`
- final net count: `247`
- exact keep-set verification passed for both paths

Additional manual validation:
- the resulting `.aedb` from the updated path was opened and inspected in SIWave
- the deleted nets were removed as expected

## Notes
- this patch intentionally does **not** change post-delete cache or refresh behavior
- this patch focuses only on the delete algorithm and preserving existing semantics

closes #2035

### DIFF
--- a/doc/changelog.d/2034.fixed.md
+++ b/doc/changelog.d/2034.fixed.md
@@ -1,0 +1,1 @@
+Improve grpc Nets.delete performance for large net deletions
--- a/src/pyedb/grpc/database/nets.py
+++ b/src/pyedb/grpc/database/nets.py
@@ -709,15 +709,32 @@ class Nets(CommonNets):
         if isinstance(netlist, str):
             netlist = [netlist]
 
-        self._pedb.modeler.delete_primitives(netlist)
-        self._pedb.padstacks.delete_padstack_instances(netlist)
+        requested_names = set(netlist)
+        if not requested_names:
+            return []
+
+        target_nets: List[Net] = []
+        primitives_to_delete: List[Any] = []
+        padstacks_to_delete: List[Any] = []
+
+        for net in self._pedb.layout.nets:
+            if net.name in requested_names:
+                target_nets.append(net)
+                primitives_to_delete.extend(list(net.core.primitives))
+                padstacks_to_delete.extend(list(net.core.padstack_instances))
+
+        for primitive in primitives_to_delete:
+            primitive.delete()
+
+        for padstack in padstacks_to_delete:
+            padstack.delete()
 
         nets_deleted = []
 
-        for i in self._pedb.nets.nets.values():
-            if i.name in netlist:
-                i.delete()
-                nets_deleted.append(i.name)
+        for net in target_nets:
+            net_name = net.name
+            net.delete()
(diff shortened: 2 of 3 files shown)


svandenb-dev (dismissed):
LGTM

svandenb-dev:
@pl170491 thanks for your help. FYI we are going to introduce a new grpc with next Service pack in june. All is done in memory and we bypass socket network traffic for local use (the only one supported). Performances are greatly improved.

svandenb-dev (approved):
LGTM

PR #2051, opened 2026-04-17
CHORE: test dotnet fix

svandenb-dev:
This PR is fixing a test forcing DotNet even with grpc.


tests/system/test_edb_components.py
```diff
@@ -293,6 +293,7 @@ def test_components_create_component_from_pins(self):
         assert len(component.pins) == 2
         edb.close(terminate_rpc_session=False)
 
+    @pytest.mark.skipif(use_grpc, reason="DotNet module not available in gRPC mode without .NET installation")
```
ecoussoux-ansys, line 296: I'd have a question regarding skipping this test in the `grpc` case: I've noticed that it seems to work fine in the CI of the other currently open PRs (grpc system tests windows and linux - for instance [this run](https://github.com/ansys/pyedb/actions/runs/24762316495)).
Is it therefore necessary to skip it?

svandenb-dev: Yes because I made a test on Linux without DotNet installed and it was failling forcing user to add DotNet.

ecoussoux-ansys (approved):

maxcapodi78 (approved):

PR #2078, opened 2026-04-23
CHORE: Add dotnet install target

SMoraisAnsys:
This PR migrates the .NET related dependencies into an install target. Now, to work with .NET, users will have to install pyedb with the target `dotnet`, e.g. `pip install pyedb[dotnet]`. Given all the work that was performed in that past weeks to make PyEDB work with gRPC and defaulting to it, this PR is another evidence confirming that this gRPC is the default approach.

Note that this will break things for users as they would have to install pyedb with this target to work with .NET.


Samuelopez-ansys:
How will impact in pyaedt? We need to add this target too somewhere?

SMoraisAnsys:
> How will impact in pyaedt? We need to add this target too somewhere?

Since we plan to use gRPC as default behavior for pyedb, it shouldn't impact users with default pyaedt settings.
If a user wants to use pyedb with .NET, then either he has the dotnet dependencies installed or he should see the warning stating to install them. Do you think about an issue to come ?

svandenb-dev (approved):
LGTM

PR #2079, opened 2026-04-24
CI: Update pyaedt testing on release

SMoraisAnsys:
Update the release process to catch missing test files and also align test configuration with what pyaedt is currently using when testing.


svandenb-dev (approved):
LGTM

Samuelopez-ansys (approved):

PR #2107, opened 2026-05-02
FIX: Rpc session fix

### LINKED ISSUE #2105: `RpcSession` will close a preexisting EDB RPC session
https://github.com/ansys/pyedb/blob/6676552037db75f8388caf713312e2d46921b3ef/src/pyedb/grpc/rpc_session.py#L64-L82

`RpcSession` uses `ansys.edb.core.session.launch_session` to start the EDB RPC Server, but this will return any preexisting session as only one is allowed by the client at a time.  Then, when its database reference count gets to zero `RpcSession` will close/stop the preexisting session, potentially resulting in an error if the code that started the session e.g. uses the session context manager and \_\_exit\_\_ tries to close it.

### LINKED ISSUE #2108: grpc shows random failures on Winddows CI

### Description of the bug

On Windows CI, the gRPC server process cleanup is slower due to socket TIME_WAIT states and process teardown delays. When close_rpc_session (class-scoped fixture) kills the server between test classes, the next class's RpcSession.start() could:
Reuse a stale rpc_session reference pointing at a dead server process
Have _SESSION_MOD.current_session still referencing the old dead session, causing __start_rpc_server to attach to it instead of starting fresh
Fail to bind the new server's port because the old socket hasn't been released yet
Fixes applied:
rpc_session.py - Health check in start(): Before reusing an existing session, validate the server process is actually alive via _is_server_alive(). If the process is dead, clean up and start fresh.
rpc_session.py - _reset_session_state(): New helper that clears all session state AND sets _SESSION_MOD.current_session = None. This prevents __start_rpc_server from falsely detecting a "preexisting" session that's actually dead.
rpc_session.py - close() hardened:
Wrapped end_managing() and disconnect() in try/except to prevent cascading failures
Added Windows-specific 1.0s delay (vs 0.1s on Linux) after disconnect to allow socket cleanup
Uses _reset_session_state() to clear all state including _SESSION_MOD.current_session
edb_init.py - Null guard in close(): Added if self._db is not None check with try/except around self._db.close() to prevent "Illegal operation on a null object" when close is called on an already-closed or failed EDB.

### Steps To Reproduce

NA

Windows

3.7

### Installed packages

NA

svandenb-dev:
This PR is adressing issue #2105.

Fixing issue #2108 
Addressing random failure on Windows CI:
On Windows CI, the gRPC server process cleanup is slower due to socket TIME_WAIT states and process teardown delays. When close_rpc_session (class-scoped fixture) kills the server between test classes, the next class's RpcSession.start() could:
Reuse a stale rpc_session reference pointing at a dead server process
Have _SESSION_MOD.current_session still referencing the old dead session, causing __start_rpc_server to attach to it instead of starting fresh
Fail to bind the new server's port because the old socket hasn't been released yet
Fixes applied:
rpc_session.py - Health check in start(): Before reusing an existing session, validate the server process is actually alive via _is_server_alive(). If the process is dead, clean up and start fresh.
rpc_session.py - _reset_session_state(): New helper that clears all session state AND sets _SESSION_MOD.current_session = None. This prevents __start_rpc_server from falsely detecting a "preexisting" session that's actually dead.
rpc_session.py - close() hardened:
Wrapped end_managing() and disconnect() in try/except to prevent cascading failures
Added Windows-specific 1.0s delay (vs 0.1s on Linux) after disconnect to allow socket cleanup
Uses _reset_session_state() to clear all state including _SESSION_MOD.current_session
edb_init.py - Null guard in close(): Added if self._db is not None check with try/except around self._db.close() to prevent "Illegal operation on a null object" when close is called on an already-closed or failed EDB.

closes #2105 
closes #2108

### DIFF
--- a/doc/changelog.d/2107.fixed.md
+++ b/doc/changelog.d/2107.fixed.md
@@ -1,0 +1,1 @@
+Rpc session fix
--- a/src/pyedb/grpc/database/padstacks.py
+++ b/src/pyedb/grpc/database/padstacks.py
@@ -208,9 +208,11 @@ class Padstacks(object):
         self.__definitions = {}
         for padstack_def in self._pedb.db.padstack_defs:
             try:
+                if padstack_def.is_null:
+                    continue
                 self.__definitions[padstack_def.name] = PadstackDef(self._pedb, padstack_def)
             except (Exception, InvalidArgumentException) as e:
-                self._logger.warning(f"Error processing padstack definition {padstack_def.name}: {e}")
+                self._logger.warning(f"Error processing padstack definition: {e}")
         return self.__definitions
 
     @property
--- a/src/pyedb/grpc/edb_init.py
+++ b/src/pyedb/grpc/edb_init.py
@@ -193,20 +193,19 @@ class EdbInit(object):
         else:
             return False
 
-    def close(self, terminate_rpc_session=None, keep_session_alive=False):
+    def close(self, terminate_rpc_session=None):
         """Close the database.
 
         Parameters
         ----------
         terminate_rpc_session : bool, optional
-            Deprecated.  Use ``keep_session_alive`` instead.
             When ``True``, forcibly terminate the RPC server regardless of how
             many other databases are still open.
-        keep_session_alive : bool, optional
-            When ``True``, the RPC server is kept running after the database is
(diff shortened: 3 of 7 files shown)


svandenb-dev:
Closing this PR as commit is included in PR #2109

PR #2119, opened 2026-05-05
FIX: rpc session start issue #2105

### LINKED ISSUE #2105: `RpcSession` will close a preexisting EDB RPC session

svandenb-dev:
This PR is trying to fix issue #2105 following the discussion with user.

closes #2105

### DIFF
--- a/doc/changelog.d/2119.fixed.md
+++ b/doc/changelog.d/2119.fixed.md
@@ -1,0 +1,1 @@
+Rpc session start issue #2105
--- a/src/pyedb/grpc/rpc_session.py
+++ b/src/pyedb/grpc/rpc_session.py
@@ -179,6 +179,21 @@ class RpcSession:
     @staticmethod
     def __start_rpc_server():
         preexisting = _SESSION_MOD.current_session is not None
+        if preexisting:
+            # A session already exists in the underlying module. Adopt it
+            # directly instead of calling launch_session which may throw or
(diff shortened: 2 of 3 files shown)


isaacwaldron:
@svandenb-dev I suggested the wrong value for `_owns_session` when `preexisting` is `True`: it should be `False`.  I've pushed a commit that results in my test on https://github.com/ansys/pyedb/tree/issue-2105 now passing.  I believe I made the correct edits to the tests you're adding here but please take a look.

svandenb-dev:
> @svandenb-dev I suggested the wrong value for `_owns_session` when `preexisting` is `True`: it should be `False`. I've pushed a commit that results in my test on https://github.com/ansys/pyedb/tree/issue-2105 now passing. I believe I made the correct edits to the tests you're adding here but please take a look.

HI @isaacwaldron thanks for your help

Samuelopez-ansys (approved):

PR #2177, opened 2026-05-19
FIX: Ruff D211 d212

hui-zhou-a:
There are conflicting ruff rules

The rule one-blank-line-before-class (D203) and blank-line-before-class (D211) are directly opposed. Same for multi-line-summary-first-line (D212) with multi-line-summary-second-line (D213).

see https://github.com/astral-sh/ruff/issues/15171

To make ruff compatible with Codacy
1, keep D203, ignore D211
2, keep D213, ignore D212


hui-zhou-a:
Close this PR as there needs to be a decsion how to deal with ruff-codacy conflicts

PR #2180, opened 2026-05-21
FIX: Unpin dependencies

### LINKED ISSUE #2179: Only pin minimum versions of runtime dependencies
As a library, PyEDB should require minimum versions of its dependencies and only if required limit the maximum version but these should not be pinned to a specific version with `==`.  As an example, currently PyEDB pins `ansys-edb-core==0.3.1` which prevents the use of pre-release versions of this library by downstream projects that depend on both PyAEDT (which currently requires `pyedb>=0.75.0`) and `ansys-edb-core`.

isaacwaldron:
Unpin dependencies to provide only minimum versions.  I have seen some discussion that this should only be applied to runtime dependencies and test/CI/CD-type dependencies should actually be pinned, but this was not the case here already so I left them as is except for unpinning any that were currently pinned.

Closes #2179

### DIFF
--- a/doc/changelog.d/2180.fixed.md
+++ b/doc/changelog.d/2180.fixed.md
@@ -1,0 +1,1 @@
+Unpin dependencies
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,8 +27,8 @@ classifiers = [
 dependencies = [
     "numpy>=1.20.0,<3",
     "pydantic>=2.6.4,<2.14",
-    "toml==0.10.2",
-    "ansys-edb-core==0.3.1",
+    "toml>=0.10.2",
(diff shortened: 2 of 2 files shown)


svandenb-dev (approved):
OK for me but waiting approval from:
@SMoraisAnsys @Samuelopez-ansys wrt PyAEDT

SMoraisAnsys (approved):
We are advising to use `>=` instead of pinning dependencies in the pyproject.toml file. For dev and CI reproducibility, you could pin down the dependencies version through a uv lock file like it's been performed in other repos.

If there is an issue within pyaedt about a new release of ansys-edb-core, then the version can be pinned on pyaedt's side.

PR #2182, opened 2026-05-21
CHORE: uv support added

### LINKED ISSUE #2186: Support uv
### Description of the feature

iv needs to be supported with pyedb

### Steps for implementing the feature

_No response_

### Useful links and references

_No response_

svandenb-dev:
This PR is adding uv installation support with pyedb.

closes #2186


.pre-commit-config.yaml
```diff
@@ -9,7 +9,9 @@ repos:
   rev: v0.15.13
   hooks:
   - id: ruff-check
+    exclude: uv\.lock
   - id: ruff-format
+    exclude: uv\.lock
```
SMoraisAnsys:
```suggestion
- id: ruff-format
```
Dunno why this is added but I assume it's an error.

.pre-commit-config.yaml
```diff
@@ -35,6 +37,7 @@ repos:
   rev: v6.0.0
   hooks:
   - id: requirements-txt-fixer
+    exclude: uv\.lock
```
SMoraisAnsys:
```suggestion
```
Same here

pyproject.toml
```diff
@@ -96,6 +96,41 @@ doc = [
 [tool.flit.module]
 name = "pyedb"
 
+# uv configuration
+[tool.uv]
+compile-bytecode = true
+
+[dependency-groups]
+dev = [
```
SMoraisAnsys, line 74: Could you update `dev` to be defined from other groups, e.g.
```
dev = [
  "prek==0.3.13",
  "ruff>=0.4.0",
  { include-group = "doc" },
  { include-group = "tests" },
]
```
where tests would contain the test related dependencies.
Also, can you remove the optional install target to only have groups for tests and docs ?

pyproject.toml
```diff
@@ -106,6 +141,7 @@ Releases = "https://github.com/ansys/pyedb/releases"
 [tool.ruff]
 line-length = 120
 fix = true
+exclude = ["uv.lock"]
```
SMoraisAnsys:
```suggestion
```
Shouldn't be required

SMoraisAnsys (dismissed):
I left a few comments and would advise to use the lock file in the CI

pyproject.toml
svandenb-dev: Done

svandenb-dev:
> I left a few comments and would advise to use the lock file in the CI

Done you can check

eblanco-ansys (dismissed):
LGTM

.github/workflows/ci-pr.yml
```diff
@@ -118,30 +118,22 @@ jobs:
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          enable-cache: false
 
       - name: Install Python library and tests dependencies
-        run: |
-          source .venv/bin/activate
-          python -m pip install .[all,tests,dotnet]
+        run: uv sync --frozen --extra all --extra tests --extra dotnet
```
SMoraisAnsys: groups should be used for tests as these dependencies should be migrated into a group
```suggestion
        run: uv sync --frozen --extra all --group tests --extra dotnet
```
Same comment for other locations

.github/workflows/ci-pr.yml
```diff
@@ -118,30 +118,22 @@ jobs:
...
 
-      - name: Create Python venv
-        run: |
-          python -m venv .venv
-
-      - name: Update pip
-        run: |
-          source .venv/bin/activate
-          python -m pip install -U pip
+      - name: Install uv
```
SMoraisAnsys, line 121: A virtual environment should be used to avoid conflicts between runs (all the more since we use self hosted runners)

.github/workflows/ci-pr.yml
```diff
@@ -118,30 +118,22 @@ jobs:
...
-          source .venv/bin/activate
-          pytest tests/unit -vvv --cov=pyedb --color=yes -ra
+          uv run --no-sync pytest tests/unit -vvv --cov=pyedb --color=yes -ra
 
       - name: Create coverage files
         run: |
-          source .venv/bin/activate
-          python -m coverage html -d .cov\unit-html
-          python -m coverage xml -o .cov\unit.xml
+          uv run --no-sync python -m coverage html -d .cov\unit-html
```
SMoraisAnsys: If a virtual env is used, there is no need to use `uv run` and `python` can be used directly

.python-version
```diff
@@ -0,0 +1,2 @@
+3.10
```
SMoraisAnsys: I wouldn't add that since we are using the minimum version of the package by default. Not a strong opinion though

pyproject.toml
SMoraisAnsys: Not done, tests is still not a groupe

SMoraisAnsys (changes requested):

.github/workflows/ci-pr.yml
eblanco-ansys: Then you mean is preferable to first activate and then use `python` normally? Both ways should work the same if I'm not wrong

.github/workflows/ci-pr.yml
SMoraisAnsys, line 484:
```suggestion
          needs-quarto: true
          group-dependencies-name: 'doc'
```
@eblanco-ansys I think this input is missing

eblanco-ansys: Thanks a lot, good catch! @SMoraisAnsys

.github/workflows/ci-pr.yml
```diff
@@ -113,24 +113,15 @@ jobs:
...
           persist-credentials: false
 
-      - name: Setup Python
-        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
+          enable-cache: false
           python-version: ${{ env.TESTS_VERSION }}
```
SMoraisAnsys, line 125: There is still no virtual env defined :/

.github/workflows/ci-pr.yml
```diff
@@ -408,12 +373,12 @@ jobs:
           retry_on: error
           timeout_minutes: 40
           command: |
-            . .venv/bin/activate
+            source .venv/bin/activate
```
SMoraisAnsys, line 379: How can that work if there is no venv defined  previously ?

.python-version
```diff
@@ -0,0 +1,2 @@
+
```
SMoraisAnsys: Remove that file if it's empty ?

pyproject.toml
```diff
@@ -62,14 +62,27 @@ all = [
...
+[dependency-groups]
 tests = [
     "mock>=5.1.0,<5.3",
     "pytest>=7.4.0,<9.1",
     "pytest-cov>=4.0.0,<7.2",
     "pytest-xdist>=3.5.0,<3.7",
-    "pytest-rerunfailures>=14.0,<17"
+    "pytest-rerunfailures>=14.0,<17",
+]
+dev = [
```
SMoraisAnsys: for visibility, can dev be defined at the end or begening of the dependency-groups section ?

SMoraisAnsys (changes requested):
Some minor comments

.github/workflows/ci-pr.yml
eblanco-ansys: Setup-python creates the venv right?

eblanco-ansys:
> Some minor comments

Pushed changes

.github/workflows/ci-pr.yml
SMoraisAnsys: No it doesn't

SMoraisAnsys (changes requested):

SMoraisAnsys (approved):

PR #2199, opened 2026-05-25
FIX: HFSS-PI simulation setup cast failing

### LINKED ISSUE #2200: HFSS-PI simulation setup cast returns always None (ansys-edb-core issue)

### Description of the bug

We can work around the issue from pyedb-core cast() method which returns always None when Simulation setup object is HFSS-Pi. 

### Steps To Reproduce

NA

Windows

3.7

### Installed packages

NA

svandenb-dev:
This PR is addressing a work around to resolve a bug in ansys-edb-core where setup.cast() returns always None for HFSS-PI simulation setup. 
This PR is also increasing code coverage for simulation setup.

closes #2200

### DIFF
--- a/src/pyedb/grpc/database/simulation_setup/hfss_pi_simulation_setup.py
+++ b/src/pyedb/grpc/database/simulation_setup/hfss_pi_simulation_setup.py
@@ -36,6 +36,8 @@ class HFSSPISimulationSetup(SimulationSetup):
 
     def __init__(self, pedb, core: "CoreHFSSPISimulationSetup"):
         super().__init__(pedb, core)
+        if not isinstance(core, CoreHFSSPISimulationSetup):
+            core = CoreHFSSPISimulationSetup(core.msg)
         self.core = core
         self._pedb = pedb
 
--- a/src/pyedb/grpc/database/simulation_setups.py
+++ b/src/pyedb/grpc/database/simulation_setups.py
@@ -24,6 +24,7 @@
 from typing import cast
 
 from ansys.edb.core.database import ProductIdType as CoreProductIdType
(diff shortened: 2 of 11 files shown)


Samuelopez-ansys (approved):

src/pyedb/grpc/database/simulation_setup/hfss_pi_simulation_setup.py
```diff
@@ -36,6 +36,8 @@ class HFSSPISimulationSetup(SimulationSetup):
 
     def __init__(self, pedb, core: "CoreHFSSPISimulationSetup"):
```
SMoraisAnsys, line 37: The signature should be updated if another value can be passed.
This applies to other setup changes

src/pyedb/grpc/database/simulation_setups.py
```diff
@@ -51,6 +52,52 @@ def __init__(self, pedb):
         self._siwave_cpa_setup: dict[str, SIWaveCPASimulationSetup] = {}
         self._hfss_pi_setups: dict[str, HFSSPISimulationSetup] = {}
 
+    def _raw_simulation_setups(self) -> list:
```
SMoraisAnsys, line 55:
```suggestion
    def _raw_simulation_setups(self) -> list[CoreSimulationSetup]:
```

svandenb-dev: Thanks @SMoraisAnsys I will add this change in another PR since this one completed almost the full CICD

src/pyedb/grpc/database/simulation_setup/hfss_pi_simulation_setup.py
svandenb-dev: I will add the it in another PR it is planned thanks

PR #2232, opened 2026-06-03
CHORE: adding grpc message

svandenb-dev:
This PR is adding grpc message and adjusting CI timeout due to tests increase.

### DIFF
--- a/src/pyedb/generic/design_types.py
+++ b/src/pyedb/generic/design_types.py
@@ -24,11 +24,16 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Literal, overload
 import warnings
 
-from pyedb.generic.grpc_warnings import GRPC_BETA_WARNING, GRPC_NOT_SUPPORTED_WARNING
+from pyedb.generic.grpc_warnings import GRPC_NOT_SUPPORTED_WARNING
 from pyedb.generic.settings import settings
 from pyedb.misc.decorators import deprecate_argument_name
 
 DEFAULT_GRPC_VERSION = 2026.1
+GRPC_DEFAULT_MESSAGE = (
+    "gRPC is now the default mode starting with Ansys release 26.1.2 (June 10th). "
+    "gRPC introduces Fast Mode, which significantly increases speed. "
+    "Make sure you are using the latest PyEDB and AEDT versions."
+)
 
 if TYPE_CHECKING:
     from pyedb.dotnet.edb import Edb as EdbDotnet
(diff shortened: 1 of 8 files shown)


Samuelopez-ansys (dismissed):

src/pyedb/generic/design_types.py
```diff
@@ -24,11 +24,16 @@
 from typing import TYPE_CHECKING, Any, Literal, overload
 import warnings
 
-from pyedb.generic.grpc_warnings import GRPC_BETA_WARNING, GRPC_NOT_SUPPORTED_WARNING
+from pyedb.generic.grpc_warnings import GRPC_NOT_SUPPORTED_WARNING
```
SMoraisAnsys, line 27: I think there is no need to keep the warning in the `grpc_warnings` module either, right ?

src/pyedb/generic/design_types.py
```diff
@@ -24,11 +24,16 @@
...
 from pyedb.generic.settings import settings
 from pyedb.misc.decorators import deprecate_argument_name
 
 DEFAULT_GRPC_VERSION = 2026.1
+GRPC_DEFAULT_MESSAGE = (
+    "gRPC is now the default mode starting with Ansys release 26.1.2 (June 10th). "
+    "gRPC introduces Fast Mode, which significantly increases speed. "
+    "Make sure you are using the latest PyEDB and AEDT versions."
+)
```
SMoraisAnsys: I think this message should be reworked, it tastes strange (very subjective here).

SMoraisAnsys (dismissed):
Don't you think it would be a good idea to wait a bit more to switch for gRPC ? Wouldn't notify users about the incoming release be a better way and then proceed with the change a bit later ?

src/pyedb/generic/design_types.py
svandenb-dev: We need to keep it for now because we still have users having Version not supported

src/pyedb/generic/design_types.py
svandenb-dev: Ok I wanted something very explicit, I agree this should be more neutral.

svandenb-dev:
I removed 2 unit tests which were failing due to the warning removal on grpc backend selection. Those tests were not valid anymore with this PR.

tests/unit/test_design_types.py
```diff
@@ -85,17 +85,6 @@ def dotnet_edb(**kwargs):
     monkeypatch.setitem(sys.modules, "pyedb.dotnet.edb", dotnet_module)
 
 
-@pytest.mark.skipif(not config["use_grpc"], reason="Applies only for grpc.")
```
SMoraisAnsys, line 88: I think some of the tests need to be updated and check that GRPC_DEFAULT_MESSAGE is used, right ?

SMoraisAnsys (approved):
LGTM, just left a minor comment

tests/unit/test_design_types.py
svandenb-dev: Yes I will do it in another PR

PR #2249, opened 2026-06-08
FEAT: issue #2248 implemented

### LINKED ISSUE #2248: Add geometry check and heal
### Description of the feature

This feature is concerning adding Siwave valcheck directly in pyedb.

### Steps for implementing the feature

_No response_

### Useful links and references

_No response_

svandenb-dev:
This PR is adding a workflow to support Siwave valcheck directly on EDB. 

closes #2248


hui-zhou-a:
@svandenb-dev there is a LayoutValidation class under Edb. It can be confusion the new method is also under Edb.

This new method uses SIwave. There are twice database conversion. The new database will not be the same as the original one. It could be better to move the new method under SIwave class to make the user aware of the database conversion.

svandenb-dev:
> @svandenb-dev there is a LayoutValidation class under Edb. It can be confusion the new method is also under Edb.
> 
> This new method uses SIwave. There are twice database conversion. The new database will not be the same as the original one. It could be better to move the new method under SIwave class to make the user aware of the database conversion.

Ah yes right I will check a way to prevent this confusion

hui-zhou-a:
> > @svandenb-dev there is a LayoutValidation class under Edb. It can be confusion the new method is also under Edb.
> > This new method uses SIwave. There are twice database conversion. The new database will not be the same as the original one. It could be better to move the new method under SIwave class to make the user aware of the database conversion.
> 
> Ah yes right I will check a way to prevent this confusion

@svandenb-dev I saw the method was moved under layout validation class. However, this mixes pyedb native vailidations and the ones from SIwave.

Samuelopez-ansys (dismissed):

Samuelopez-ansys (approved):

PR #2254, opened 2026-06-09
REFACTOR: remove redundant methods in gRPC

hui-zhou-a:
The removed methods do exist in dotnet. They are never exposed to customers before gRPC, so they should not exist.


svandenb-dev (approved):
LGTM

PR #2295, opened 2026-06-19
CHORE: Release ci adjusted

svandenb-dev:
This PR is adjusting Ci release to enable PyEDB release publishing due to SP2 bug introduced in AEDT / PyAEDT.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -531,7 +531,7 @@ jobs:
         run: |
           . .venv\Scripts\activate
           pytest -v external/pyaedt/tests/system/solvers/test_analyze.py
-        timeout-minutes: 180
+        timeout-minutes: 240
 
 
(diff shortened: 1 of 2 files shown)


maxcapodi78 (approved):

SMoraisAnsys (changes requested):
I think this version is not working with pyaedt and it doesn't seem to be related with SP2.
There is currently a run failing while using pyaedt VM (not leveraging the SP2) see https://github.com/ansys/pyaedt/actions/runs/27832295961/job/82371803620?pr=7802

SMoraisAnsys (approved):
The changes seem to pass correctly in pyaedt VM's (with the current main version of pyedb) so I don't expect any breaks with the release. Approving.

See https://github.com/ansys/pyaedt/actions/runs/27833856097/job/82376993404?pr=7802

PR #2306, opened 2026-06-22
FIX: cutout bugs

maxcapodi78:
- components pins was including also Terminals
- Setups was failing to retrieve setups if some non-valid setup was present (CPA)
- some connectable was causing retrievial of information
- Multipin resistors where not handled into extended_net.py


svandenb-dev (approved):
LGTM

PR #2311, opened 2026-06-23
FIX: various improvements in plot method

maxcapodi78:
Ability to change the color and size of outline
ability to plot components on top and bottom
ability to change the color of components and labels


svandenb-dev (approved):
LGTM

PR #2408, opened 2026-07-19
FEAT: trace.create width value support added

### LINKED ISSUE #2392: `Path.create` hints `width` as `float` but underlying API accepts more types
https://github.com/ansys/pyedb/blob/635ecc2c862c7cb9c9b324f855d70317946191fc/src/pyedb/grpc/database/modeler.py#L509

Since 0.2.0 ansys-edb-core has supported `ValueLike` for the `width` argument of [Path.create](https://edb.core.docs.pyansys.com/version/stable/api/_autosummary/ansys.edb.core.primitive.path.Path.create.html#ansys.edb.core.primitive.path.Path.create).

svandenb-dev:
This PR is adding Value support with Path.create()

closes #2392


src/pyedb/grpc/database/primitive/path.py
```diff
@@ -58,9 +58,9 @@ def width(self) -> float:
         Returns
         -------
         float
-            Path width or None.
+            Path width in meters.
         """
-        return Value(self.core.width)
+        return float(Value(self.core.width))
```
hui-zhou-a: Value is a child class of float. Do you need to convert it again?

svandenb-dev: Some test are failing because returning the Value which is float but still Value class makes internal api failing sometimes (One test is failing on CI if I remove explicit conversion) that's why I keep it.

hui-zhou-a: @svandenb-dev  which test is failing due to Value? If you convert it to real float, the user is not able to retreive variable name.

src/pyedb/grpc/database/primitive/path.py
```diff
@@ -52,22 +52,22 @@ def __init__(self, pedb, core=None):
...
-    def width(self) -> float:
+    def width(self) -> float | int:
         """Path width.
 
         Returns
         -------
         float
             Path width in meters.
         """
-        return float(Value(self.core.width))
```
hui-zhou-a: @svandenb-dev it should return Value object

hui-zhou-a (dismissed):

src/pyedb/grpc/database/primitive/path.py
svandenb-dev: Done

hui-zhou-a (approved):

PR #2437, opened 2026-07-27
CHORE: vulnerability-security cleanup

svandenb-dev:
This PR is cleaning up some try/except exception handling.

### DIFF
--- a/src/pyedb/grpc/rpc_session.py
+++ b/src/pyedb/grpc/rpc_session.py
@@ -190,7 +190,12 @@ class RpcSession:
         Returns
         -------
         bool
-            ``True`` when the connection was established successfully, ``False`` otherwise.
+            ``True`` when the connection was established successfully.
+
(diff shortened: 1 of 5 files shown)


src/pyedb/grpc/rpc_session.py
```diff
@@ -234,10 +231,7 @@ def connect_to_existing_server(port, ip_address="localhost"):
             session_obj.connect()
         except Exception as exc:
             settings.logger.error(f"Failed to connect to RPC server at {ip_address}:{port}: {exc}")
```
hui-zhou-a: @svandenb-dev  is it correct to log error rather than raising error?

svandenb-dev: Ok I changed, it raises an error now rather than a log.

hui-zhou-a (approved):

PR #2543, opened 2026-09-08
FIX: group location fixed

### LINKED ISSUE #2542: Group.location bad format -< returning tuple with x as Value instead of float

### Description of the bug

Component.location is returning bad tuple format.

### Steps To Reproduce

NA

Windows

3.7

### Installed packages

NA

svandenb-dev:
This PR is addressing issue #2542 with fixing bad location tuple format. It was returning tuple (x, y) with X as Value and y as float.

closes #2542

### DIFF
--- a/src/pyedb/grpc/database/hierarchy/group.py
+++ b/src/pyedb/grpc/database/hierarchy/group.py
@@ -99,12 +99,12 @@ class Group:
 
         Returns
         -------
-        List
-            [x, y].
+        tuple[float, float]
+            (x, y) coordinates.
 
         """
         location = self.core.location
-        return self._pedb.value(location[0]), location[1].value
+        return location[0].value, location[1].value
 
     @location.setter
     def location(self, value: tuple | list):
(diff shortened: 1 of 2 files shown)


src/pyedb/grpc/database/hierarchy/group.py
```diff
@@ -99,12 +99,12 @@ def location(self) -> tuple:
...
         Returns
         -------
-        List
-            [x, y].
+        tuple[float, float]
+            (x, y) coordinates.
 
         """
         location = self.core.location
-        return self._pedb.value(location[0]), location[1].value
```
hui-zhou-a, line 107: Is it corret to return core value instead of pyedb value?

svandenb-dev: .value is a float not Value object. That's the equivalent to ToDouble() from DotNet

hui-zhou-a: How to return the parameter name if location is parameterized

svandenb-dev: Ok I changed to pyedb Value

hui-zhou-a (approved):

PR #2545, opened 2026-09-11
FIX: Primitive queries fix

### LINKED ISSUE #2546: edb.layout.filter method does not return all primitives with grpc backend

### Description of the bug

grpc backend has a bug concerning primitive filtering. Especially on gds based EDB, some primitives are grouped while DotNet list all. decompose method must be used to provide same behavior as DotNet.

### Steps To Reproduce

NA

Windows

3.7

### Installed packages

NA

svandenb-dev:
This PR is addressing issue #2546. layout.primitive filtering does not account all primitives with grpc backend especially with gds based edb where primitives are grouped. 

closes #2546


tests/conftest.py
```diff
@@ -172,6 +172,15 @@ def get_package(self, edbapp=True, additional_files_folders="", version=None):
         else:
             return target_file
 
+    def get_edb_from_gds(self, edbapp=True, additional_files_folders="", version=None):
```
hui-zhou-a: why is this method called get_edb_from_gds

svandenb-dev: I removed this method

src/pyedb/grpc/database/layout/layout.py
```diff
@@ -363,8 +364,39 @@ def primitives_by_aedt_name(self) -> dict[str, Primitive]:
...
+        underlying EDB database.
+        """
+        if self.__collections_expanded:
+            return
+        self.__collections_expanded = True
+
+        collections = []
+        for primitive in self.core.primitives:
+            if _resolve_primitive_type_name(primitive) == "PrimitiveInstanceCollection":
+                collections.append(primitive)
```
hui-zhou-a, line 387: Why is decomponsition needed? primitives from the PrimitiveInstanceCollection don't belong to layout anymore. You are changing edb data structure.

hui-zhou-a: How do I get all primitives don't belong to any PrimitiveInstanceCollection?

svandenb-dev: No that's only with grpc new feature. When an EDB is loaded from GDS some primitive are assigned into collection, this is not related to LayoutInstance. Dev made it like this to bring gds based lighter but we need to use .decompose method to get some primitive count Vs DotNet

src/pyedb/grpc/database/layout/layout.py
```diff
@@ -289,12 +291,26 @@ def filter_primitives(
...
+            lazily: it only runs once, the first time it is requested.
 
         Returns
         -------
         list[Primitive]
             Filtered primitives.
         """
+        if expand_instance_collections:
+            self.expand_primitive_instance_collections()
+
```
hui-zhou-a, line 314: find_primitive should only return primitives. Nothing else. However, when expand_instance_collections=True, instanceCollection is decomposited. The database is changed.

src/pyedb/grpc/database/layout/layout.py
```diff
@@ -363,6 +379,39 @@ def primitives_by_aedt_name(self) -> dict[str, Primitive]:
...
+        directly) since decomposition **mutates** the underlying EDB database and can be
+        expensive for large collections. It is only performed once per layout (lazily cached).
+        """
+        if self.__collections_expanded:
+            return
+        self.__collections_expanded = True
+
+        collections = []
+        for primitive in self.core.primitives:
+            if _resolve_primitive_type_name(primitive) == "PrimitiveInstanceCollection":
```
hui-zhou-a, line 405: Could you make a new property def primitive_instance_collections?

src/pyedb/grpc/database/layout/layout.py
svandenb-dev: Ok I will change the behavior then

hui-zhou-a (approved):

